Issue #95, opened 2020-09-14
Create UI Gatsby theme


unamaria:
Should include UI components, like a page wrapper (with needed providers like helmet’s) and common ones like icon or form elements. Takes care of its own configuration and stuff like declaring modules for assets (if needed). Added by default in Gatsby starter.

- Add any React components that are reusable across projects from the existing starter
- Add any configuration needed for it to work properly with the starter
- Include from starter
- Come up with a sane name/way to clearly reference it

pmelab:
We where looking at themes, but most of it is doable with simple react component composition. And since themes are a static inheritance mechanism, they are less flexible.

PR #100, opened 2020-09-14
feat: extract gatsby theme from starter

marcotisi:
## Description

This PR:
- creates a basic `@amazeelabs/gatsby-amazee-theme`
- configures `tailwind` and `post-css` with the same plugins of https://github.com/AmazeeLabs/gatsby-starter-amazee
- removes the configuration for those packages from `silverback-website` and `gatsby-starter`
- uses the new theme in `silverback-website`
- adds `Header`, `Footer` and `SEO` components to `gatsby-theme-core` as long as `useSiteMetadata` hook and a `Layout`
- shadows `Footer` component in `silverback-website` to demonstrate how shadowing works.

## Motivation

We need to extract common plugins and settings from `gatsby-starter` and `silverback-website` into a new `gatsby-amazee-theme`

## Related issue

#95 

## Tested

No

## Notes

Please note that `tailwind.config.js` and `tailwind.css` are still scaffolded from `gatsby-starter` because those two files are always changed when creating a new website.


marcotisi:
@pmelab @unamaria what do you think about the PR description? I think that it should give a good context to a reader who doesn't know the details and, if you like it, we could create a PR template inside this monorepo.

pmelab:
> @pmelab @unamaria what do you think about the PR description? I think that it should give a good context to a reader who doesn't know the details and, if you like it, we could create a PR template inside this monorepo.

Yes, a PR template is a good idea. 👍

pmelab (changes requested):
Should we rename the theme to `gatsby-theme`? In analogy to `gatsby-starter`? We don't need the `amazee` since its already in our namespace. Or perhaps `gatsby-theme-base` since there might be more?

And I think we should prototype the full workflow before merging this in to make sure everything works:

- [x] add a component to `silverback-theme` that is directly used in `silverback-website`
- [x] add a component to `silverback-theme` that is overridden in `silverback-website`
- [ ] integrate storybook for `silverback-theme` (directly in there? separate package?)
- [ ] integrate storybook for `silverback-website` (maybe look into [Storybook Composition](https://medium.com/storybookjs/storybook-composition-af0da9084fba)?)

marcotisi:
Yes, sounds fine to me! I think that we could integrate Storybook in a separate PR, just to avoid having too many irons in the fire. What do you think?

pmelab (approved):

PR #226, opened 2020-10-05
chore(drupal/cypress): added phpstan checks

Leksat:
## Package(s) involved
drupal/cypress

## Description of changes
Added PHPStan checks for Cypress module.

In the end I used https://github.com/mglaman/phpstan-drupal. But first I tried to integrate PHPStan and Drupal myself... It was a nightmare 😅

## Motivation and context
Philipp.

## How has this been tested?
Automatically 😎

### DIFF
--- a/apps/silverback-drupal/composer.json
+++ b/apps/silverback-drupal/composer.json
@@ -104,6 +104,7 @@
         ]
     },
     "require-dev": {
+        "jangregor/phpstan-prophecy": "^0.8.0",
         "mglaman/phpstan-drupal": "^0.12.5",
         "phpstan/extension-installer": "^1.0"
     }
--- a/apps/silverback-drupal/composer.json
+++ b/apps/silverback-drupal/composer.json
@@ -102,5 +102,9 @@
             "if [ -d web/modules/custom ]; then phpunit web/modules/custom; fi",
             "yarn install && CYPRESS_TAGS=@COMPLETED cypress run"
         ]
+    },
+    "require-dev": {
+        "mglaman/phpstan-drupal": "^0.12.5",
+        "phpstan/extension-installer": "^1.0"
     }
 }
--- a/packages/composer/drupal/cypress/phpstan.neon
+++ b/packages/composer/drupal/cypress/phpstan.neon
@@ -1,0 +1,15 @@
+parameters:
+    level: 8
+    checkGenericClassInNonGenericObjectType: false
+    checkMissingIterableValueType: false
+    paths:
+        - cypress.module
+        - cypress.install
+        - src
+        - tests
+        - scripts
+    ignoreErrors:
+        # Maybe we should enable it at some point.
+        - '#\Drupal calls should be avoided in classes, use dependency injection instead#'
+        # That's Drupal requirement to use `new static()` and we can do nothing about it at the moment.
+        - '#\Unsafe usage of new static\(\).#'
--- a/packages/composer/drupal/cypress/src/CachedInstallation.php
+++ b/packages/composer/drupal/cypress/src/CachedInstallation.php
@@ -107,13 +107,13 @@ class CachedInstallation {
    *
    * @param string $appRoot
    *   The path to the Drupal root directory.
-   * @param $siteDir
+   * @param string $siteDir
    *   The site directory to use.
-   * @param $lockId
+   * @param string $lockId
    *   The lock id used for this test site install.
-   * @param $dbUrl.
+   * @param string $dbUrl.
    *   The database url
-   * @param $dbPrefix
+   * @param string $dbPrefix
    *   The database prefix to use.
    */
   public function __construct($appRoot, $siteDir, $lockId, $dbUrl, $dbPrefix) {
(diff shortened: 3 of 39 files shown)


apps/silverback-drupal/composer.json
```diff
@@ -102,5 +102,9 @@
             "if [ -d web/modules/custom ]; then phpunit web/modules/custom; fi",
             "yarn install && CYPRESS_TAGS=@COMPLETED cypress run"
         ]
+    },
+    "require-dev": {
+        "mglaman/phpstan-drupal": "^0.12.5",
+        "phpstan/extension-installer": "^1.0"
```
Leksat, line 109: This can be removed once `drupal/cypress` is released and the above dependencies are visible for composer from the package metadata.
(Composer loads the package from a local directory, but it reads the list of dependencies from the Packagist and there is no way to force it to read metadata from the local package `composer.json`)

Leksat:
@pmelab you can find the list of issues found by PHPStan in the failed Test job. Short summary:
- 2 cases of `Unsafe usage of new static()` (I don't quite understand what is wrong with that and how it can be fixed)
- 6 cases of `\Drupal calls should be avoided in classes, use dependency injection instead`
- `Undefined variable: $suites` (actually, potentially undefined)

packages/composer/drupal/cypress/phpstan.neon
```diff
@@ -0,0 +1,15 @@
...
+    checkMissingIterableValueType: false
+    paths:
+        - cypress.module
+        - cypress.install
+        - src
+        - tests
+        - scripts
+    ignoreErrors:
+        # Maybe we should enable it at some point.
+        - '#\Drupal calls should be avoided in classes, use dependency injection instead#'
```
Leksat, line 13: Maybe...

packages/composer/drupal/cypress/src/CachedInstallation.php
```diff
@@ -410,14 +439,19 @@ protected function restoreCache($cacheDir) {
...
     $files = [];
     $finder = new Finder();
     $finder->files()->in($cacheDir);
     $finder->ignoreDotFiles(FALSE);
     foreach ($finder as $file) {
-      $files[$file->getRelativePath() . '/' . $file->getBasename()] = $file->getRealPath();
+      $realpath = $file->getRealPath();
+      if ($realpath !== FALSE) {
+        $files[$file->getRelativePath() . '/' . $file->getBasename()] = $realpath;
+      }
```
Leksat, line 454: At first the code was
```php
      if ($realpath === FALSE) {
        throw new \Exception('Cannot get realpath for a file: "' . $file->getRelativePath() . '/' . $file->getBasename() . '".');
      }
      $files[$file->getRelativePath() . '/' . $file->getBasename()] = $realpath;
```
But then I've got `Exception: Cannot get realpath for a file: "/settings.php".`

So I changed it to the current less strict version.

Not sure if it's right though.

Leksat:
I have fixed 200+ issues found by PHPStan... after I switched the level from 1 to 8 😅

Leksat:
I tried the https://github.com/phpstan/phpstan-strict-rules plugin, but it's too strict I think. Attaching its output: [strict-rules.txt](https://github.com/AmazeeLabs/silverback-mono/files/5341888/strict-rules.txt)

Leksat:
Output from https://github.com/phpstan/phpstan-deprecation-rules:
```
$ vendor/bin/phpstan analyse --configuration=web/modules/contrib/cypress/phpstan.neon
 41/41 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ------ ----------------------------------------------------------------------------------------------------------------------------------------------------------------
  Line   /Users/alex/work/silverback-mono/apps/silverback-drupal/web/modules/contrib/cypress/src/EventSubscriber.php
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------------------------
  31     Parameter $event of method Drupal\cypress\EventSubscriber::onRequest() has typehint with deprecated class Symfony\Component\HttpKernel\Event\GetResponseEvent:
         since Symfony 4.3, use RequestEvent instead
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------------------------

 ------ -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  Line   /Users/alex/work/silverback-mono/apps/silverback-drupal/web/modules/contrib/cypress/src/EventSubscriber/RemoveErrorHeadersSubscriber.php
 ------ -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  24     Parameter $event of method Drupal\cypress\EventSubscriber\RemoveErrorHeadersSubscriber::onResponse() has typehint with deprecated class Symfony\Component\HttpKernel\Event\FilterResponseEvent:
         since Symfony 4.3, use ResponseEvent instead
 ------ -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

 ------ ---------------------------------------------------------------------------------------------------------------
  Line   /Users/alex/work/silverback-mono/apps/silverback-drupal/web/modules/contrib/cypress/src/TtyProcessManager.php
 ------ ---------------------------------------------------------------------------------------------------------------
  18     Call to deprecated method inheritEnvironmentVariables() of class Symfony\Component\Process\Process:
         since Symfony 4.4, env variables are always inherited
 ------ ---------------------------------------------------------------------------------------------------------------


 [ERROR] Found 3 errors
```

Since we want the module to work with the older Drupal versions which may use the older dependencies, I don't think it makes sense to fix deprecation warnings upfront. It will be super-hard to track what is safe to fix and what is not.

Maybe let's wait until the deprecated things are removed? Then we will have no choice 😄

pmelab (approved):

PR #248, opened 2020-10-07
Gatsby schema snapshots

pmelab:
## Package(s) involved
* silverback-website
* @amazeelabs/gatsby-starter

## Description of changes
Added the `gatsby-plugin-schema-snapshot`. It will create a snapshot of the Gatsby schema that is stored in the git repository and is only updated when `yarn schema-update` is specifically run.

## Motivation and context
When using Drupal with JSON:API or Contentful, Gatsby infers the schema automatically based on the data it gets. If a field is not populated anywhere, it won't show up in the result set and since these sources don't expose typing information, Gatsby won't put it into the schema. This causes the build process to break if a query requests this field.

Using this plugin, the schema is locked so that can't happen. Also, by committing the schema snapshot to git, it's possible to run type checks without having to build the project, which saves some CI time.

The downside is that there are two files now, that cause conflicts. But they only change when the Gatsby schema changes, and not during regular frontend development. And the conflict can be resolved easily by running `yarn schema-update`.

### DIFF
--- a/apps/silverback-website/.gitignore
+++ b/apps/silverback-website/.gitignore
@@ -70,5 +70,4 @@ yarn-error.log
 .yarn-integrity
 
 # Ignore generated files
-generated/*
-!generated/.gitkeep
+generated/types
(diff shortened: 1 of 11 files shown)


marcotisi (approved):
Good job, this way we can control when to generate schema snapshots 👍

PR #273, opened 2020-10-12
chore(amazeelabs/php-dev): php-dev composer package

Leksat:


Leksat:
This won't work. I mean we can't have a single dev package for all of the projects.

Case:
- we don't want strict rules in Cypress module
- we want strict rules on other module
- we add the strict rules package to php-dev
- it is automatically used for Cypress module, which we do not want

I'll create amazeelabs/cypress-dev package.

PR #274, opened 2020-10-12
chore(drupal/cypress-dev): dev dependencies for drupal/cypress

Leksat:


Leksat:
@pmelab I created a separate `drupal/cypress-dev` package instead of a common `amazeelabs/php-dev`. Details are here: https://github.com/AmazeeLabs/silverback-mono/pull/273#issuecomment-706981161

PR #368, opened 2020-10-27
Extract react-di to a public package

### LINKED ISSUE #334: Extract react dependency injection to package
The silverback website contains a minimalistic dependency injection mechanism for react. This should be extracted to its own package (e.g. `@amazeelabs/react-di`) and can serve as an example for #328 .

https://github.com/AmazeeLabs/silverback-mono/tree/development/apps/silverback-website/src/utils

Leksat:
Closes #334


packages/npm/@amazeelabs/react-di/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "dependencies": {},
+  "devDependencies": {
+    "@amazeelabs/jest-preset-react": "^1.1.0",
+    "typescript": "^4.0.5"
+  },
+  "scripts": {
+    "test:ci": "jest",
+    "test:watch": "jest --watch",
+    "test": "is-ci test:ci test:watch"
+  }
```
pmelab, line 22: We will also need a build step that produces ES6 in for example a `dist` folder. This will work in the monorepo, but when using in another project, `node_modules` are most of the time not transpiled.

Example: https://medium.com/cameron-nokes/the-30-second-guide-to-publishing-a-typescript-package-to-npm-89d93ff7bccd

pmelab (changes requested):

packages/npm/@amazeelabs/react-di/package.json
pmelab: Also, the package is usable without typescript. No sane person would do that, but 🤷

Leksat: Something like this? https://github.com/AmazeeLabs/silverback-mono/compare/b6077c9..bbe982b

pmelab: Perfect, thanks!

pmelab (approved):

Issue #481, opened 2020-11-17
silverback-drupal app is not compatible with composer 2


JohnAlbin:
When using composer 2, the `composer install` command fails with:

```
wikimedia/composer-merge-plugin dev-master requires composer-plugin-api ^1.1 -> found composer-plugin-api[2.0.0] but it does not match the constraint.
```

Upstream bug report: https://github.com/wikimedia/composer-merge-plugin/issues/184

Since we are doing `composer install` as a postinstall task of the Drupal app and `yarn install` causes an install of both Drupal and Gatsby apps, this means if you are trying to use the Gatsby starter app, the `yarn install` command fails because the app you don't want to use (Drupal) can't be installed properly.

2 possible fixes:

* Remove the `"scripts": { "postinstall": "composer install" },` lines from `apps/silverback-drupal/package.json`.
* Remove the `"wikimedia/composer-merge-plugin": "^1.4"` line from `packages/composer/amazeelabs/silverback-cli/composer.json`. We can add it back later when it is compatible with Composer 2.

…or we can do both.

JohnAlbin:
I'm leaning towards doing both of these things.

1. If you are installing the Drupal app, most would already know you need to run `composer install` and we can add it to that app's README for those that don't know to run it.
2. If you are installing the Gatsby app, you don't want a composer command taking up install time and disk space.
3. If we just did the first option (remove postintall), that still doesn't fix the composer 2 problem.

JohnAlbin:
We have an additional problem to fixing composer 2 compatibility. The `apps/silverback-drupal/composer.lock` file has locked `amazeelabs/silverback-cli` to an older version that doesn't support Composer 2.

The `apps/sliverback-drupal/composer.lock` shows:

```
        {
            "name": "amazeelabs/silverback-cli",
            "version": "dev-drupal-cypress",
            "dist": {
                "type": "path",
                "url": "../../packages/composer/amazeelabs/silverback-cli",
                "reference": "75ee1e347d8abf06b0ae102f28013620509202ce"
            },
            "require": {
                "alchemy/zippy": "dev-master",
                "cweagans/composer-patches": "^1.6.7",
                "drush/drush": "^10.3.4",
                "ext-zip": "*",
                "vlucas/phpdotenv": "^5.1",
                "wikimedia/composer-merge-plugin": "^1.4"
            },
```

[...]

PR #482, opened 2020-11-17
fix(drupal): add Composer 2 compatibility

JohnAlbin:
## Package(s) involved
amazeelabs/sliverback-cli

## Description of changes

See #481


pmelab:
I'm not 100% sure about removing `composer.lock`. In real applications it's definitely required, and for the sake of automated module testing, I wonder if it needs to be there and auto-updated by renovatebot.

Leksat:
Fixed in #500 and #502.

PR #490, opened 2020-11-25
Fix silverback-website builds

Leksat:
## Problem
TypeScript and Webpack fail to resolve `@amazeelabs/gatsby-theme-core` and `@amazeelabs/react-di` when they are not compiled.

## Solution
Compile the mentioned packages during the monorepo initialization.

## Solution which did not work
https://github.com/microsoft/TypeScript/issues/21423 suggest a workaround:
```ts
// index.ts in root of each package
export * from './src/index'
```

It helps TypeScript, but does not work for webpack.

### DIFF
--- a/apps/silverback-website/package.json
+++ b/apps/silverback-website/package.json
@@ -60,7 +60,7 @@
     "clean": "gatsby clean",
     "update-schema": "GATSBY_UPDATE_SCHEMA_SNAPSHOT=true gatsby build",
     "codegen": "graphql-codegen --config codegen.yml",
-    "test:ci": "jest && yarn build",
+    "test:ci": "jest && yarn codegen && yarn tsc && yarn build",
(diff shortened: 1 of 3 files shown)


Leksat:
@pmelab FYI: I tried a workaround but it did not work. See the PR description.

Leksat:
@pmelab also I'm not sure if the newly added `postinstall` scripts will fire when packages are installed from NPM 🤔

pmelab:
@Leksat `npm` triggers the `prepare` script on local install as well as before publishing. We should make sure everything that has to be built does this on `prepare`.

Leksat:
@pmelab yes, npm does that. But yarn does not. It does not support all of npm hooks: https://yarnpkg.com/advanced/lifecycle-scripts

So the issue with local development is still open.

Leksat:
How about this? #494

PR #494, opened 2020-11-30
Prepare TS packages for local usage

Leksat:


pmelab:
Yep, that's good. I would just limit it to one hook and make sure all packages use this one. Since `prepare` is the correct solution for `npm` and we solve it in `yarn` by running it on `postinstall`, I would suggest going for that.

Leksat:
@pmelab better now?

pmelab:
👌 Perfect, thanks!

PR #505, opened 2020-12-01
Cypress: trigger a test

Leksat:
to check if cypress test fails not because of composer 2


Leksat:
Hmm, it's actually because of the Composer 2 🤔

PR #506, opened 2020-12-01
CI caching and test dependencies

Leksat:


Leksat:
No cache:
- pre cache: 0
- yarn install: 4m 25s
- post cache: 50s

With cache:
- pre cache: 20s
- yarn install: 2m 4s
- post cache: 24s

The difference isn't that big, but we have no issues with caching.

Leksat:
The same, but with yarn.lock in git #510

No cache:
- pre cache: 0
- yarn install: 4m 16s
- post cache: 55s

With cache: 
- pre cache: 24s
- yarn install: 2m 18s
- post cache: 28s

So for the CI it does not really matter if we keep yarn.lock in git.

PR #509, opened 2020-12-01
build: clear root lockfile before installing

pmelab:
I ran into problems with multiple versions of `react`, caused by outdated versions of `react` in the root lock file. Deleting the lock file and reinstalling fixed it. Should we just always clear it when running `yarn` since its ephemeral anyway?

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "commit": "yarn git-cz",
     "test": "yarn lerna run prepublish --since --include-dependencies && yarn lerna run test --since --stream --concurrency=1",
     "git-registry": "yarn verdaccio -c config.yml",
+    "preinstall": "rm -f yarn.lock",
     "postinstall": "lerna bootstrap && lerna run prepare"
   },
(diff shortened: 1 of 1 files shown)


Leksat:
Local tests.

If there is no yarn.lock (it is removed in both preinstall and postinstall hooks):
- each root `yarn install` take 1m 30s
- a `yarn add something` in an app takes 1m

If there is yarn.lock (without this PR):
- each root `yarn install` takes from 14s to 40s
- a `yarn add something` in an app takes 10s

If yarn.lock "blinks" (this PR):
- each root `yarn install` takes 1m 30s
- a `yarn add something` in an app takes 10s

Leksat:
@pmelab according to my test above, this PR changes nothing if `yarn add` is used in an app or a package. You still may run into conflicts.

If we also remove the lock file from postinstall, it makes work with yarn really slow.

How about an alternative approach? We keep things as they are, but we recommend to remove yarn.lock locally in case of conflicts.

PR #512, opened 2020-12-02
Test Cypress on Drupal updates

Leksat:


Leksat:
Current error:
```
  cy.drupalScript
    1) executes a php script to create a node


  1 passing (21s)
  1 failing

  1) cy.drupalScript
       executes a php script to create a node:
     CypressError: `cy.request()` failed on:

http://localhost:8888/cypress/script

The response we received from your web server was:

  > 500: Internal Server Error
```

PR #517, opened 2020-12-03
--force should not ignore existing config

Leksat:
From the `--force` flag help:
> Force installation.

The actual action was: Force installation and ignore the existing config.


Leksat:
@pmelab I need your opinion here. It could be that I'm not getting something.

Leksat:
This change will be a part of a bigger PR (coming soon!)

PR #524, opened 2020-12-07
Remake Gatsby build status check

### LINKED ISSUE #521: Drupal/Gatsby: Build monitoring
Based on the prototype already implemented by @Leksat 

Gatsby cloud does not allow to register other routes (like `__is_refreshing` in the current prototype). So we could try to store the state in Drupal and send HTTP updates from the build process.

Also, I think we should move this to a dedicated module, that can be installed along with the existing `gatsby` module (or even become a part of it eventually).

- [x] Implement a `gatsby_build_monitor` drupal module that manages a single state (idle, building, failed) for preview _and_ production builds, provides HTTP endpoints for setting these states and uses these to show it in the toolbar.
- [x] Implement a `gatsby-plugin-build-monitor` gatsby plugin, that uses the logic from the current prototype to send build status information to an HTTP endpoint.

Leksat:
A part of #521.


pmelab (changes requested):
This seems to work also on Gatsby Cloud! I think we can move on with this solution:

- [ ] Rephrase the status in the frontend: `idle` -> `Preview ready`, `rebuilding` -> `Preview updating`
- [ ] Add a second state for the production build.
- [ ] Add a documentation page.
- [ ] Move Drupal module and Gatsby plugin to `packages` and publish them to drupal.org/npm.

Leksat:
As we have learned, it did not work on the Gatsby Cloud.

In the last call with Philipp we decided to postpone (and remove) the Preview Refresh indicator. Instead I will make the Site Rebuild indicator.

PR #556, opened 2021-01-07
Make Silverback CLI use drupal-scaffold

Leksat:
The idea is to maintain as less as possible scaffold files.

Changes:
- Most of assets are removed. All Lagoon files can be brought via [amazeeio/drupal_integrations](https://packagist.org/packages/amazeeio/drupal_integrations) package.
- Assets which survived:
  - `.envrc`
  - `.env.local`: renamed from `.env` to avoid conflicts with Lagoon and `amazeeio/drupal_integrations`.
  - `settings.php`:
    - now it's only an addition to original `settings.php`
    - it also works if `settings.php`is scaffolded (e.g. by `amazeeio/drupal_integrations`)
  - `.gitignore`: just a small addition
- `silverback init` is removed.
- Most of env vars are removed.
- The project is now tested with both `drupal/recommended-project` template and [amazeeio/drupal-example-simple](https://github.com/amazeeio/drupal-example-simple) starter (which is based on `amazeeio/drupal_integrations`).

In theory it should work with _any_ Drupal-Composer project relying on `drupal/core-composer-scaffold`. This is why the setup instructions are bigger than before.


Leksat:
@pmelab I would also remove `bin/silverback-dev` because its purpose isn't clear to me. But maybe you still use it 😛

pmelab (approved):

PR #568, opened 2021-01-19
docs(silverback-website): add Formik documentation

mattiasimonato:
## Package(s) involved
`silverback-website`

## Description of changes
This PR adds the missing Formik documentation. It focuses on the basics, fields validation, Custom fields

## Motivation and context
We need to motivate the choice of choosing Formik and how to use it.

## Screenshot

<details>
  <summary>Click to expand!</summary>
 
formik-docs-preview
</details>

### DIFF
--- a/apps/silverback-website/docs/gatsby/formik.mdx
+++ b/apps/silverback-website/docs/gatsby/formik.mdx
@@ -57,35 +57,35 @@ export function FormikText({
 Custom field usage in a formik form example:
 
 ```jsx
-import React from "react";
-import ReactDOM from "react-dom";
-import { Formik, Field, Form } from "formik";
-import { FormikText } from "./FormikText"
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { Formik, Field, Form } from 'formik';
+import { FormikText } from './FormikText';
 
 function App() {
   return (
     <div className="App">
       <h1>Contact Us</h1>
       <Formik
-        initialValues={{ name: "", email: "" }}
-        onSubmit={async values => {
-          await new Promise(resolve => setTimeout(resolve, 500));
+        initialValues={{ name: '', email: '' }}
+        onSubmit={async (values) => {
+          await new Promise((resolve) => setTimeout(resolve, 500));
           alert(JSON.stringify(values, null, 2));
         }}
       >
         <Form>
-	  <Field
-	    name="name"
-	    label="name"
-	    type="text"
-	    component={FormikText} //<--- how to use your custom field
-	  />
-	  <Field
-	    name="email"
-	    label="email"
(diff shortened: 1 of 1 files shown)


apps/silverback-website/docs/gatsby/formik.mdx
```diff
@@ -0,0 +1,99 @@
...
+Custom field (that handles label, input type and errors) example:
+
+```jsx
+import React from "react"
+
+export function FormikText({
+  field, // { name, value, onChange, onBlur }
+  form: { touched, errors }, // also values, setXXXX, handleXXXX, dirty, isValid, status, etc.
+  ...props
+}) {
```
marcotisi: Since we almost use arrow functions in silverback, I think it's better to also have examples which use them:

```suggestion
export const FormikText = ({
  field, // { name, value, onChange, onBlur }
  form: { touched, errors }, // also values, setXXXX, handleXXXX, dirty, isValid, status, etc.
  ...props
}) => {
```

apps/silverback-website/docs/gatsby/formik.mdx
```diff
@@ -0,0 +1,99 @@
...
+
+Custom field usage in a formik form example:
+
+```jsx
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { Formik, Field, Form } from 'formik';
+import { FormikText } from './FormikText';
+
+function App() {
```
marcotisi: Same as above:
```suggestion
export const App = () => {
```

marcotisi (changes requested):
I only have a small suggestion about using arrow functions instead of regular functions.

pmelab:
@mattiasimonato perfect! 👍 

We should also add Formik and Yup (are there other helpful libraries?) right to the gatsby starter. And maybe we can even come up with some minimal use case, so we have a working example that we can test against updates.

apps/silverback-website/docs/gatsby/formik.mdx
mattiasimonato: Updated 👍

apps/silverback-website/docs/gatsby/formik.mdx
mattiasimonato: Updated 👍

mattiasimonato:
@pmelab thanks for the review!
A form example for testing is a really good idea, I would keep it as simple as possible!
For the gatsby starter, I think we can add only **formik** and **yup** at the moment 👍

marcotisi (approved):
I also agree with @pmelab, a minimum example would be helpful.
But let's keep the PR simple and merge, you can then open a new one using another scope to add the minimum example and the required setup.

PR #573, opened 2021-01-25
JS-Tooling cleanup

pmelab:
Instead of having multiple versions for jest and ESLint configs, they now inspect the packages dependencies to decide if certain features (Storyshots, React ...) are added. The packages @amaazeelabs/{eslint-config,jest-preset} therefore already include react-specifics now, so the *-react variants are not needed any more.

I also added the `@amazeelabs/scaffold` package that will install all necessary tools and configs in a package. The idea is _not_ to commit the configuration, but rather add it on `postinstall` so we can easily adjust them across packages.

For a more detailed rundown have a look at the `test.sh` file in `@amazeelabs/scaffold`.


packages/npm/@amazeelabs/scaffold/.gitignore
```diff
@@ -0,0 +1,4 @@
+dist
+node_modules
+yarn-error.log
+test
```
Leksat, line 4: What's `test` for? Local experiments?

UPD: found it!

packages/npm/@amazeelabs/scaffold/README.md
```diff
@@ -0,0 +1,24 @@
...
+## Installation
+
+```shell
+yarn add @amazeelabs/scaffold
+yarn amazee-scaffold
+```
+
+This will alter the packages `.gitignore` file, add a `postinstall` hook and
+make sure that configurations for code quality tools like `eslint`, `prettier`
+or `jest` are in place. The scaffolded configuration files should are added to
```
Leksat: Typo: should _be_ added

packages/npm/@amazeelabs/scaffold/README.md
```diff
@@ -0,0 +1,24 @@
...
+`.gitignore` and should not be committed. They will be auto-added and updated
+with every `yarn install`.
```
Leksat, line 17: If I already have dependencies installed, and then I git pull and run `yarn install` again, will this run the `postinstall` hook and update the scaffolded files?

packages/npm/@amazeelabs/scaffold/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Usage
+
+After install there is a `test` npm script that will run tests either in CI-mode
+or in watch-mode, depending on the current environment. It will also look for an
+optional `test.sh` in the package root that will also be executed when running
+in an CI environment.
```
Leksat, line 24: `test.sh` is not very intuitive for me. Also I personally do not like when local and CI commands work too differently. But I always can run `CI=true yarn test`, so all good.
(Just thoughts 😄 )

packages/npm/@amazeelabs/scaffold/test.sh
```diff
@@ -0,0 +1,53 @@
...
+
+# Initiate a new node package.
+yarn init -y
+
+# Install the scaffold package.
+yarn add ../
+
+# Run the initial scaffold to install dotfiles and scripts.
+yarn amazee-scaffold
+yarn husky install && rm -rf .husky
```
Leksat: So husky should isn't managed by amazee-scaffold?

packages/npm/@amazeelabs/scaffold/test.sh
```diff
@@ -0,0 +1,53 @@
...
+if [ -f jest.config.js ]; then
+  echo "Git repository not properly cleaned up."
+  exit 1
+fi
+
+# Run yarn install which should re-scaffold files.
+yarn
+
+if [[ ! -f "jest.config.js" ]]; then
+  echo "jest.config.js was not scaffolded"
```
Leksat: Trick I learned recently to echo to stderr:
```
>&2 echo ...
```
I like that when I read the script it emphasizes that this is an error message.

Leksat (approved):
I left few optional notes/questions.

Looks awesome in general!

packages/npm/@amazeelabs/scaffold/.gitignore
Leksat: Still, shouldn't `test` be  added to `@amazeelabs/scaffold`s `.gitignore`? And not to all packages?

packages/npm/@amazeelabs/scaffold/README.md
pmelab: Yes, that's the idea.

packages/npm/@amazeelabs/scaffold/README.md
pmelab: You can always run `yarn test:ci` too. And the "local" runs the "watchable" tests (jest, eslint).
But they diverged a lot, I see that. Maybe we even remove `is-ci-cli`, have a `test` command and another `watch` command to run during development?

packages/npm/@amazeelabs/scaffold/test.sh
pmelab: I did not fully understand why we have to run `yarn husky install` again 😬 
Do you have an idea how we could manage it by `amazee-scaffold`?

packages/npm/@amazeelabs/scaffold/README.md
Leksat: Yes, would be great!

packages/npm/@amazeelabs/scaffold/test.sh
Leksat: Make `yarn amazee-scaffold` run `yarn husky install && rm -rf .husky` for us?

packages/npm/@amazeelabs/scaffold/.gitignore
pmelab: `.gitignore` is not copied over. it just edits the projects `.gitignore` to ignore the scaffolded files.

Leksat: Ahhh!

packages/npm/@amazeelabs/jest-preset/package.json
```diff
@@ -9,16 +9,14 @@
   "publishConfig": {
     "access": "public"
   },
-  "bin": {
-    "jest-preset-configure": "config/configure.js"
+  "peerDependencies": {
+    "@storybook/addon-storyshots": "^6.1.9"
```
marcotisi, line 13: I would not add `@storybook/addon-storyshots` to the list of peer dependencies because it's not really needed, the preset only test if the package is installed and then add the storybook preset.

marcotisi (commented):
I've added a note about `peerDependencies`. I see that you're basically using this package as a source for scaffolding packages, scripts, and other files. Why not creating a `stubs` folder, add there the source files to be merged/copied over so we could avoid ignoring specific parts that are only used to make this package works?
That way we could also simplify the scripts and the `cli.js` would look like this
```js
#!/usr/bin/env node
require('esm')(module /*, options*/)('./dist').scaffold(
  path.resolve(path.join(__dirname, 'stubs')),
  process.cwd(),
);
```

What do you think?

pmelab:
> I've added a note about `peerDependencies`. I see that you're basically using this package as a source for scaffolding packages, scripts, and other files. Why not creating a `stubs` folder, add there the source files to be merged/copied over so we could avoid ignoring specific parts that are only used to make this package works?
> That way we could also simplify the scripts and the `cli.js` would look like this
> 
> ```js
> #!/usr/bin/env node
> require('esm')(module /*, options*/)('./dist').scaffold(
>   path.resolve(path.join(__dirname, 'stubs')),
>   process.cwd(),
> );
> ```
> 
> What do you think?

My idea was to make this package "eat it's own dog food", so we can get immediate feedback on changes and don't have to duplicate the config for itself.

But maybe we could have a stubs folder, and the package scaffolds itself on install from it 🤯

marcotisi:
Yeah, I agree with you, but having a separate folder it's more explicit and less prone to mistakes I guess. Then we can completely remove these lines and the code that manages them.
```ts
/**
 * Packages that are used by @amazeelabs/scaffold that are _not_ installed
 * in scaffolded projects.
 */
export const ignoredPackages = ['chalk', 'esm'];

/**
 * Files in @amazeelabs/scaffold that are _not_ installed in scaffolded projects.
 */
export const ignoredFiles = [
  'cli.js',
  'index.js',
  '.gitignore',
  'package.json',
  'test.sh',
  'README.md',
];
```
Also, if in the future a developer adds a new dependency that it's only required by the package itself and forgets to add it to the list of ignored packages, we would end up with an unneeded dependency on the final project.

pmelab:
@marcotisi I implemented your suggestions. Would you have another look?

packages/npm/@amazeelabs/scaffold/.gitignore
```diff
@@ -0,0 +1,14 @@
...
+
+### MANAGED BY @amazeelabs/scaffold - START
+.eslintrc.js
+.huskyrc
+.jest-runner-eslintrc
+.lintstagedrc
+.prettierrc
+jest.config.js
+tsconfig.json
+### MANAGED BY @amazeelabs/scaffold - END
```
marcotisi, line 14: Is this intended or is the package trying to add ignores to itself?

marcotisi (approved):
Seems fine to me, I have only one doubt about the package scaffolding itself 😆 .
I'll approve it anyway, feel free to go ahead if that's not the case.

packages/npm/@amazeelabs/scaffold/.gitignore
pmelab: This is intended _and_ the package is managing its own ignores 😄 Thats the "eat your own dogfood" part.

marcotisi: Cool! Let's get it merged then!

PR #592, opened 2021-02-25
Add frontend tech stack documentation

marcotisi:
## Package(s) involved
- silverback-website

## Description of changes
This PR adds the documentation regarding the frontend tech stack,

## Motivation and context
We need a list of battle-tested packages that we can use in our projects, without worrying if they work or not.

## Screenshot
<details>
  <summary>Click to expand!</summary>

localhost_8000_gatsby_tech-stack

</details>

### DIFF
--- a/apps/silverback-website/docs/gatsby/packages.mdx
+++ b/apps/silverback-website/docs/gatsby/packages.mdx
@@ -1,0 +1,243 @@
+---
+title: Tech stack
+path: /gatsby/tech-stack
+nav: Gatsby
+position: 7
+---
+
+# Tech stack
+
+Decide which package to use for a specific functionality is often hard. You need
+to take into account a lot of different aspect, like:
+
+- is it easy to use for other developers?
+- is it reliable?
+- is it actively maintained?
+- is it well documented?
+
+Here you'll find a list of battle-tested libraries that have already been used on our
+projects. But remember, no package is universal, and you might end up using a
+new one. Just ask yourself the questions above, and if the answer is yes, open a
+pull request to add the new library to this documentation, so we can have a
+discussion about integrating it in our stack.
+
+You can use these websites to get different indicators, like popularity,
+quality, stars or download count. Is then up to you to pick the right one for
+your purpose.
+
+- [npm](https://www.npmjs.com/)
+- [Openbase](https://openbase.com/)
+
+## Fetching data
+
+### [`axios`](https://github.com/axios/axios)
+
+**Description**: Promise based HTTP client for the browser and node.js.
+
+**Features**:
+- Make XMLHttpRequests from the browser
+- Make http requests from node.js
+- Supports the Promise API
+- Intercept request and response
+- Transform request and response data
+- Cancel requests
+- Automatic transforms for JSON data
+- Client side support for protecting against XSRF
+
+### [`react-query`](https://github.com/tannerlinsley/react-query)
+
+**Description**: Hooks for fetching, caching and updating asynchronous data in
+React.
+
+**Features**:
+
+- Transport/protocol/backend agnostic data fetching (REST, GraphQL, promises,
+  whatever!)
+- Auto Caching + Refetching (stale-while-revalidate, Window Refocus,
+  Polling/Realtime)
+- Parallel + Dependent Queries
+- Mutations + Reactive Query Refetching
+- Multi-layer Cache + Automatic Garbage Collection
+- Paginated + Cursor-based Queries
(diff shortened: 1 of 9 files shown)


pmelab:
I think we should split this list into "defaults" we somehow install upfront and "on-demand" packages. The defaults actually *should* be used instead of cooking up own stuff (e.g `classnames`, `date-fns`), while others might be added use case specific (`axios`, `swiper`)

apps/silverback-website/docs/gatsby/packages.mdx
```diff
@@ -0,0 +1,243 @@
...
+You can use these websites to get different indicators, like popularity,
+quality, stars or download count. Is then up to you to pick the right one for
+your purpose.
+
+- [npm](https://www.npmjs.com/)
+- [Openbase](https://openbase.com/)
+
+## Fetching data
+
+### [`axios`](https://github.com/axios/axios)
```
pmelab: Whats the benefit of `axios` over `fetch` + polyfills? Also we have made good experience with [GraphQL Codegen](https://graphql-code-generator.com/) and auto-generating typed API's, that essentially take away all direct contact with the fetching library.  Maybe it would be worth looking into [OpenAPI Generator](https://github.com/OpenAPITools/openapi-generator) for REST endpoints.

apps/silverback-website/docs/gatsby/packages.mdx
```diff
@@ -0,0 +1,243 @@
...
+  Polling/Realtime)
+- Parallel + Dependent Queries
+- Mutations + Reactive Query Refetching
+- Multi-layer Cache + Automatic Garbage Collection
+- Paginated + Cursor-based Queries
+- Load-More + Infinite Scroll Queries w/ Scroll Recovery
+- Request Cancellation
+- React Suspense + Fetch-As-You-Render Query Prefetching
+
+### [`jsonapi-react`](https://github.com/aribouius/jsonapi-react)
```
pmelab: Not sure if we should promote that, since we are pretty sure that we don't default to JSON:API. Rather standardise on `react-query` + generated API's from GraphQL Codegen or OpenAPI Codegen.

apps/silverback-website/docs/gatsby/packages.mdx
```diff
@@ -0,0 +1,243 @@
...
+- Auto-saving forms
+- Dependent Validation
+- Conditional Logic
+- Dynamic Fields
+- React Native
+- Tree shakeable
+
+## Maps
+
+### [`google-map-react`](https://github.com/google-map-react/google-map-react)
```
pmelab: This might be something that needs a little additional research. A lot of projects, especially in government, can't use Google Maps for legal reasons. I would be good to have an alternative. Currently I see two other options:

* Fully open: Leaflet + OpenStreetMap (https://react-leaflet.js.org/)
* Paid service, but not Google: MapBox + react-map-gl (https://visgl.github.io/react-map-gl/)

Largely depends on the actual requirements, so we should come up with a comparison + decision guideline.

pmelab (commented):
At this point I would consider everything thats in _Misc_ as a must. Nobody should ever run regexes against query strings, dates or html entities. Also `classnames` should be the gold standard for conditionally applying classes, especially in combination with Tailwind's purge.

Other libraries that come into my mind that are worth considering:

* [GraphQL codegen](https://graphql-code-generator.com/) (as mentioned above)
* [Headless UI](https://github.com/tailwindlabs/headlessui/tree/develop/packages/%40headlessui-react)
* Internationalisation Library (research topic)
* Responsive Images solution (when gatsby-preprocessing is not an option - also research topic)

apps/silverback-website/docs/gatsby/packages.mdx
marcotisi: Yeah, I'd rather prefer to add additional libraries when needed, so we don't put too much effort into researching libraries before we actually make use of them.

apps/silverback-website/docs/gatsby/packages.mdx
marcotisi: Yeah, that makes sense. I'm gonna remove `jsonapi-react` from the list.

apps/silverback-website/docs/gatsby/packages.mdx
marcotisi: Actually, the benefits are very few. Let's get rid of that in favor of auto-generated typed APIs.

pmelab (approved):

PR #595, opened 2021-02-26
Drupal module rules and recommendations

pmelab:
A proposal on how to deal with contributed Drupal modules and a list of "confirmed" modules. Please review the rules as well as this first list of modules, before we add more detailed explanation to them.

### DIFF
--- a/apps/silverback-website/docs/drupal/modules.mdx
+++ b/apps/silverback-website/docs/drupal/modules.mdx
@@ -23,7 +23,7 @@ unstable dependencies.
 
 ## Confirmed modules
 
-### GraphQL version 4
+### [GraphQL version 4](https://www.drupal.org/project/graphql)
 
 GraphQL is the default solution for exposing data and operations to the
 frontend. Every project should have its own custom GraphQL schema that documents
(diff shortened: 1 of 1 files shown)


apps/silverback-website/docs/drupal/modules.mdx
```diff
@@ -0,0 +1,72 @@
...
+## Prohibited modules
+Don't even bother to start a discussion.
+
+### Paragraphs
+Paragraphs generate immense load on the database and lead to lacking (up to non-
+existent) content modeling. Compared to other page building solutions, the
+editing experience is abysmal, it produces a high level of confusion and
+complexity in multilingual setups and should be avoided at all cost.
+
+### JSON:API
```
fjgarlin, line 98: I know I'm not supposed to start a discussion here but... sometimes json-api might be more than enough to cover the client or project needs, so banning beforehand feels too radical.

Projects needs and budget should probably determine this. We know we prefer GraphQL but JSON-API could still be a choice I believe.

pmelab: This was exactly my opinion, but the last months working JSON:API with changed that. If we consume the API ourselves (Gatsby or other frontends), we hurt us by not having the flexibility of GraphQL, and when creating a public API, exposing all the Drupalisms is not a good idea anyway.

Recent experience has shown that GraphQL is the cheaper option, and no matter the Budget, each project should spend the least possible amount of money.

apps/silverback-website/docs/drupal/modules.mdx
```diff
@@ -0,0 +1,72 @@
...
+unstable dependencies.
+
+
+## Confirmed modules
+
+### GraphQL version 4
+
+GraphQL is the default solution for exposing data and operations to the
+frontend. Every project should have its own custom GraphQL schema that documents
+the domain model and acts as a contract been Drupal and all consumers.
```
dan2k3k4:
```suggestion
the domain model and acts as a contract between Drupal and all consumers.
```

apps/silverback-website/docs/drupal/modules.mdx
```diff
@@ -0,0 +1,72 @@
...
+For in-depth on how to create custom GraphQL schemas, please refer to the
```
dan2k3k4:
```suggestion
For an in-depth overview on how to create custom GraphQL schemas, please refer to the
```

dan2k3k4 (changes requested):
Fix two typos in the GraphQL text description

apps/silverback-website/docs/drupal/modules.mdx
Leksat: @fjgarlin is it an often case when you need to expose a lot (or almost everything) of Drupal's data?

My experience with GraphQL v4 on Gatsby-Drupal prototype:
- 1.5 hours to learn how to write resolvers and custom data producers and implement the first of them.
- About 6 hours in total to implement the whole schema and update the frontend code (GraphQL v3 was used before).

Of course the [schema was small](https://github.com/AmazeeLabs/silverback-mono/blob/f4edd625cac4f49f0df53a45e1616505d28a506f/apps/silverback-drupal/web/modules/custom/silverback_gatsby/graphql/silverback_gatsby.graphqls), but I spent not that much of time implementing it and the frontend DX was significantly improved.

Leksat (approved):
Fine enough for a start I think.

apps/silverback-website/docs/drupal/modules.mdx
fjgarlin: Thinking on the cases where we don't need to expose a lot of data, like a small extension where part of the site is exposed.

My point is that, whilst it might not be a suitable solution for most cases, it might still be a good fit for some, where the requirements or integrations aren't too complex.

In any case, this is more maintenance point of view (ie: extension) rather than building a project from scratch, so all your points above are still valid I guess and it's no big deal to leave it out from the `silverback-mono` part. 

If any particular project really needs json-api, it will go only for that project.

Leksat: @fjgarlin could you please share some projects where json-api worked well? I'm curious to see how it is used there.

fjgarlin: Hey @Leksat, for one of our clients, they just needed to export certain area of the site (shop related) via an API so that data would be consumed by another team. They didn't have any specific needs on endpoints nor format of the data, so we just enabled the json-api modules and configured/exported everything.

Also, thinking on the cases where data will be consumed by other systems (not controlled by us), where those systems will just need endpoints and expected data, as opposed to needing to build a query/mutation.

My point from the beginning was (and is) to not prohibit it as it can have suitable use cases. I'm happy with GraphQL being the preferred option.

PR #610, opened 2021-03-08
Prevent Drupal updates config loss

Leksat:
Problem:
When Renovate updates Drupal core/modules, there might be update hooks which change the config. So it's possible that when we install Drupal with `silverback-cli`:
- `drush updb` is executed => config is updated by the update hooks
- `drush cim` is executed => all new config changes are lost

Solution:
- During release we update the Drupal install cache. So it is always fresh in the `main` branch.
- On Renovate PRs we run an additional check which
  - Installs Drupal with `--no-config-import` (yet, the update hooks are still executed)
  - Exports Drupal config
  - Fails the job if there are config changes


.github/workflows/release.yml
```diff
@@ -73,6 +73,40 @@ jobs:
       - name: Run tests
         run: yarn test
 
+      - name: Update Drupal install cache
```
pmelab, line 76: This is something we want to reuse I think. Maybe we could event turn it into a Github action in the future.

.github/workflows/renovate-checks.yml
```diff
@@ -0,0 +1,80 @@
...
+          set -e
+          cd apps/silverback-drupal
+          source .envrc
+          composer i
+          silverback setup --no-config-import
+          drush cex -y
+          cd -
+          if [[ $(git status --porcelain -- apps/silverback-drupal/config) ]]
+          then
+            >&2 echo 'Error: Found uncommitted Drupal config. Please checkout this branch locally, run `silverback setup --no-config-import && drush cex -y` and commit the config changes.'
```
pmelab, line 75: I think this solution is perfect for now. Just out of curiosity: does an auto-commit not work or did you just not invest time there?

apps/silverback-drupal/config/sync/core.extension.yml
```diff
@@ -5,8 +5,8 @@ module:
   block: 0
   breakpoint: 0
   ckeditor: 0
-  content_sync: 0
   dblog: 0
+  default_content: 0
```
pmelab, line 9: Could you state the reasons why you switched from `content_sync` to `default_content`? For future reference and not repeating ourselves 😄

pmelab (approved):

.github/workflows/renovate-checks.yml
Leksat: I haven't try. In the Release job we just commit and hope that lerna will do the push. But here we would need a commit and a push.

I can take a look at that later. If the whole thing works as expected.

pmelab: Yes, there is no rush. The manual resolution is also fine. And if it happens often, we can automate it.

apps/silverback-drupal/config/sync/core.extension.yml
Leksat: `content_sync`:
- no D9 support yet, I had to patch it hard (btw, I just recall to remove the patch 🙂 https://github.com/AmazeeLabs/silverback-mono/pull/610/commits/434176160c2f4949c6538dc4ecc78fb385e2997f)
- not really maintained
- when you run it, there can be errors and warnings in the output, yet it still somehow woks
- you have to run a command to import the content
- there was an issue with importing content after drupal install cache update

`default_content`:
- was recommended to me by @colorfield, he had good experience with it
- actively maintained by Berdir
- works more stable
- you don't need to run additional commands, content is added when you install a module
- no issues with drupal install cache update

I think it can be added to our recommended modules list.

.github/workflows/renovate-checks.yml
Leksat: I'd leave it manual forever:
- config updates are rare case (I think 😅)
- if they happen, it might make sense to review the changes

PR #633, opened 2021-04-06
internationalization

mattiasimonato:
## Package(s) involved
[react-intl](https://github.com/AmazeeLabs/react-Internationalization/tree/main/packages/react-intl)

## Motivation and context
the goal is to find a standard solution for Gatsby & React internationalization

full research [here](https://github.com/AmazeeLabs/react-Internationalization)


marcotisi (approved):
Looks good to me 👍  . I've moved the library under the `Default` sections, as almost every website we create is multilingual.

PR #638, opened 2021-04-14
Add prettier for PHP

pmelab:
## Package(s) involved
* `@amazeelabs/prettier-config`
* `amazee/silverback-cli` (as the first lab rat)

## Description of changes
Add and configure the official php plugin for prettier (https://github.com/prettier/plugin-php). Reformatted source code for `silverback-cli` to test it.

## Motivation and context
I love prettier, and I hate formatting php files.
It's not 100% Drupal standard, but the joy of nicely formatted php on save outweighs that for me.

### DIFF
--- a/packages/composer/amazeelabs/silverback-cli/.prettierrc
+++ b/packages/composer/amazeelabs/silverback-cli/.prettierrc
@@ -1,0 +1,1 @@
+"@amazeelabs/prettier-config"
--- a/packages/composer/amazeelabs/silverback-cli/package.json
+++ b/packages/composer/amazeelabs/silverback-cli/package.json
@@ -11,5 +11,11 @@
     "registry": "http://localhost:4873",
     "repository": "git@github.com:AmazeeLabs/silverback-cli.git",
     "branch": "master"
+  },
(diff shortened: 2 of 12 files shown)


fjgarlin:
Whilst I love the idea of adding this from the beginning and format things properly straight away, the fact that in conflicts (slightly) with Drupal and that we will mostly use this for Drupal code makes it a questionable candidate for me.

It can actually be seen in this PR that it will affect a good number of places in standard Drupal code.

Also, they don't seem keen on adding options that could make it compatible with Drupal, as seen here: https://github.com/prettier/plugin-php/issues/1555 

In there they suggest to use this: https://marketplace.visualstudio.com/items?itemName=ValeryanM.vscode-phpsab which will make it compatible with Drupal and its coding standards.

If the result is equally good, shouldn't we aim for the latter instead?

pmelab:
I'm not sure they are completely equal. I think `phpcbf` does not reformat e.g. argument lists that get too long? I also could be wrong.

Nonetheless, you are right. We get at least _most_ out of `phpcbf`. I wanted to try prettier-php badly 😄

Issue #200, opened 2020-09-26
Contribution managment


pmelab:
We need a documented, and as automated as possible, workflow for contributing to external dependencies. The primary focus should be pull requests.

- [ ] maintain modified versions of dependencies
- [ ] submit pull requests
- [ ] request forked versions for projects that automatically update patches
- [ ] run tests of patched packages

pmelab:
has been finally solved by #690

PR #644, opened 2021-04-26
Managed forks

pmelab:
## Description of changes
Added a workflow to managed patched packages. There is a central `forks.json` that maintains a path within silverback-mono with another repository. On release, whatever is in this path (in the first example the patched version of drupal/gutenberg) is pushed to the repository. Other projects then can simply apply the patched version directly and will receive updates throught renovatebot.

## Motivation and context
We want to centrally manage patches and patched versions of Drupal modules, so patches don't have to be maintained on every project separately.

## Proposed workflow

When a Drupal module has to be patched, it is added (along with the patches) to `silverback-drupal` and ideally covered with some tests. We also create a separate repository, prefixed with `fork-` (like https://github.com/AmazeeLabs/fork-gutenberg). Then the path of the patched module in `silverback-drupal` as well as the repository are added to `forks.json`. From there CI takes over and will push a new commit whenever our fork changes somehow. Either because the base package was updated by renovate, or the applied patches change. The fork then can simply be published under the `amazeelabs` namespace on packagist. Client projects can require the module via `composer require amazeelabs/fork-gutenberg` and receive updates from there.

## How does it work?

1. `silverback-drupal` requires a module and applies patches (standard composer workflow)
2. the release process of this repository clones the "target repository" and commits all content from within the module (which has already been patched)
3. the target repository is automatically updated, and therefore the packagist mirror
4. client projects only require the mirror (e.g. `amazeelabs/fork-gutenberg`)

Related issue: #200

### DIFF
--- a/forks.js
+++ b/forks.js
@@ -1,0 +1,40 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const path = require('path');
+const { execSync } = require('child_process');
+
+const forks = JSON.parse(fs.readFileSync(path.resolve(__dirname, 'forks.json')).toString());
+const tmpDir = path.resolve(__dirname, '.forks-tmp');
+
+if (fs.existsSync(tmpDir)) {
+  execSync(`rm -rf ${tmpDir}`);
+}
+
+fs.mkdirSync(tmpDir);
+
+process.chdir(tmpDir);
+
+forks.map(fork => {
+  const dir = new URL(fork.repository).pathname.substr(1);
+  execSync(`mkdir -p AmazeeLabs`);
+  execSync(`mkdir -p tmp/AmazeeLabs`);
+  execSync(`git clone ${fork.repository} tmp/${dir}`);
+  execSync(`cp -R ../${fork.path} ${dir}`);
+  execSync(`rm -rf ${dir}/.git`);
+  execSync(`cp -R tmp/${dir}/.git ${dir}/.git`)
+  process.chdir(path.resolve(tmpDir, dir));
+  const result = execSync(`git status --porcelain`).toString();
+  console.log(result);
+  if (result.trim().length > 0) {
+    execSync(`git add -f *`);
+    fs.readdirSync(path.resolve(tmpDir, dir)).map(file => {
+      if (file.substr(0, 1) === '.') {
+        execSync(`git add -f ${file}`);
+      }
+    })
+    execSync(`git commit -m "chore: updated fork"`);
+    execSync(`git push`);
+  }
+  process.chdir(tmpDir);
+});
--- a/forks.js
+++ b/forks.js
@@ -18,11 +18,19 @@ process.chdir(tmpDir);
 const forks = JSON.parse(fs.readFileSync(path.resolve(__dirname, 'forks.json')).toString());
 
 forks.map(fork => {
-  const dir = new URL(fork.repository).pathname.substr(1);
-  execSync(`mkdir -p AmazeeLabs`);
-  execSync(`mkdir -p tmp/AmazeeLabs`);
+  const repoUrl = new URL(fork.repository);
+  // If this runs in a github action, we have to add credentials to the remote
+  // url.
+  if (process.env.GITHUB_ACTOR && process.env.GITHUB_TOKEN) {
+    repoUrl.username = process.env.GITHUB_ACTOR;
+    repoUrl.password = process.env.GITHUB_TOKEN;
+  }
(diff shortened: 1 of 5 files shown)


fjgarlin:
@pmelab - probably we should mention somewhere how the fork is generated, the patches file (ie: https://github.com/AmazeeLabs/fork-gutenberg/blob/main/PATCHES.txt), etc. 

For example, I'm not sure how you got that fork and where the patches files live.

forks.js
```diff
@@ -0,0 +1,40 @@
...
+
+fs.mkdirSync(tmpDir);
+
+process.chdir(tmpDir);
+
+forks.map(fork => {
+  const dir = new URL(fork.repository).pathname.substr(1);
+  execSync(`mkdir -p AmazeeLabs`);
+  execSync(`mkdir -p tmp/AmazeeLabs`);
+  execSync(`git clone ${fork.repository} tmp/${dir}`);
```
fjgarlin: We might need branching support, to target the original branches (ie: `8.x` vs `1.x`...)

pmelab: That should already be handled, since the required branch is defined in `composer.json` in `silverback-drupal`.

rodrigoaguilera: I think the drawback is that this hides the upstream version that is in use on a client project and forces us to use.
```
    "minimum-stability": "dev",
```
To require the fork. A future improvement might be to raise some kind minor release version every time something changes on the fork so is more obvious when you are updating.

rodrigoaguilera:
I don't like that requiring a fork might make us more comfortable with the "patched-upstream" situation and not encourage so much to eventually remove the patch but I can see the benefits for a module used in multiple sites not having to resolve the patching conflicts many times.

If I understand correctly this not yet the forking as the silverback template should require the fork.

I think we can go ahead with the approach and see how it unfolds as multiple projects start using. If we don't like it seems easily reversible by requiring the upstream again and applying the patches.

rodrigoaguilera (approved):
@fjgarlin In the case of the PATCHES.txt linked the reference to the patch has the issue queue number but just something people experienced with Drupal know.

We need to enforce better comments for patches so is easy to find the issue upstream and eventually fix it.

fjgarlin:
@rodrigoaguilera - it's more about the file itself, not the contents (which are 100% ok). How was that file (auto?)generated so people know how to create forks.

forks.js
pmelab: I also thought about that. But i don't see a way to create a "sub-version". If we just increase the patch-version, it will collide with the next patch version of the module. If we run our own versioning, we loose all version-range requirements other dependencies have.

But can't we keep global `"minimum-stability": "dev"` and just use `-dev` version constraints when requiring the fork? We can just get every commit there, since every commit of the fork was initiated by a release of the origin.

pmelab:
Added a workflow documentation page.

Leksat:
So with this solution we have some effort involved:
  - Supporting the solution itself.
  - Adding new patched packages. 
  - Reacting to Renovate updates.
 
Wouldn't it make sense to put this effort into pushing patches to upstream?

From my experience, usually maintainers do not mind if someone fixes a bug. They only wait for RTBC. And this is where it stops most of the time.

Several times I tried to ask for help in #drupal-contributions. It worked very few times. But I realize that it is very hard for someone to find time and dive into an issue you have no clue about.

Maybe instead of managed forks, we could look into direction of collaborative work on issues?

I have no great ideas here, but maybe we can create a Github project just for managing issues where each issue is linked to the original package issue. Because
- Drupal modules are allowed to use their own issue trackers, so we can't use d.o as a central place.
- Drupal.org issue tracker is hard.
- We will be able to see who's working on what.

Then we can spend one hour per week to look there and work on something. This is not much, but I think even with this little time all the "dead" patches would be committed to upstream long ago.

Again, I think that the problem of "stuck" patches exists only because it is always just _one_ person working on an issue. We can't RTBC our own patches.

(And for the hard cases, when a maintainer do not agree with what we do, isn't it better/easier to go with a new competitive module?)

pmelab:
Summary of a discussion with an @Leksat and @fjgarlin :

We also try the "proxy-package" approach, where the package in the `amazeelabs` namespace dose not hold the full patched source, but rather a dependency to the original package along with the patches that are applied. This should reduce complexity, although there was a reason why we did not do this in the first place. We are about to find out.

pmelab:
Added the alternative approach in #690

pmelab:
Abandoned in favor of #690

Issue #450, opened 2020-07-09
Component folder structure/file naming


pmelab:
More a question / request for documentation:

Why do we create folders for components with `index.tsx`/`index.stories.tsx` etc instead of skipping the folder and naming the file according to the component directly? Multiple open tabs called `index.tsx` always confuse me.

pmelab:
@marcotisi We already talked about recent experience. What is your suggestion?

marcotisi:
We adopted the Atomic Design pattern, so my suggestion is to use a flat folder structure, with a single `PascalCase` file per component.
Stories, mocks and tests are placed respectively inside a `__stories__`, `__mocks__` and `__tests__` folders nested at the same level.
Hooks and utilities have a dedicated folder and use the `camelCase` notation.
Every folder, except for `__stories__`, `__mocks__` and `__tests__` contain an `index.ts` file used to export everything from that folder.
Everything else uses lowercase slugged notation.
So, the example above would be written like this:

```
src
├── components
│   ├── 0-atoms
│   │   ├── __mocks__
│   │   │   ├── HeroTitle.mock.tsx
│   │   ├── __stories__
│   │   │   ├── HeroTitle.stories.tsx
│   │   ├── __tests__
│   │   │   ├── HeroTitle.test.tsx
│   │   ├── HeroTitle.tsx
│   │   ├── index.ts
│   ├── 1-molecules
│   │   ├── ...
│   ├── hooks
│   │   ├── useHeroTitle.tsx
│   │   ├── index.ts
│   ├── utils
│   │   ├── __tests__
│   │   │   ├── buildUrl.test.ts
│   │   ├── buildUrl.ts
│   │   ├── index.ts
│   ├── templates
│   │   ├── blog-page.tsx
│   ├── pages
│   │   ├── about-us.tsx
```

What do you think @pmelab @dan2k3k4  ?

pmelab:
That sounds reasonable to me. The only thing I'm not a fan of are number-indexed folders (`0-atoms` ...), since these will show up in imports and cause mayhem should we ever have to adjust the order. Alphabetical ordering will put `layouts` at the wrong place, but in storybook they can be sorted correctly, and in source code its not a critical issue i think.
I would rather move everything thats not in storybook (`hooks`, `utils`) out of `components`. But maybe thats just a mistake in your "ascii art".

marcotisi:
Removing number-indexed folders sounds fine for me 👍 

[...]

PR #651, opened 2021-05-04
Storybook tailwind UI recipe

mattiasimonato:
## Description of changes
recipe that create UI structure with React, Storybook and tailwind

### DIFF
--- a/packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md
+++ b/packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md
@@ -1,1 +1,13 @@
 # Add storybook UI and Tailwind
+
+```typescript
+// Check node version.
+$$('node -v', {
+  stdout: $$.minimalVersion('12'),
+});
+
+// Check yarn version.
+$$('yarn -v', {
+  stdout: $$.minimalVersion('1.0'),
+});
+```
--- a/packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md
+++ b/packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md
@@ -199,9 +199,11 @@ $$.chdir(`layouts`);
 
 ```typescript
 # |-> StandardLayout.tsx
-import React from 'react';
+import React, {PropsWithChildren} from 'react';
 
-export const StandardLayout: React.FC = ({ children }) => (
+type Props = PropsWithChildren<{}>;
+
+export const StandardLayout = ({ children }: Props) => (
   <div className="max-w-7xl mx-auto">
     <header className="text-4xl p-4 bg-blue-50">Header</header>
     <main className="p-4 prose">{children}</main>
--- a/packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md
+++ b/packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md
@@ -160,3 +160,56 @@ export const parameters = {
   },
 };
 ```
+
+## Stories creation
+
+Delete default storybook stories
+
+```typescript
+$$.chdir(`../`);
+$$('rm -fr stories/');
+$$(`mkdir -p stories/assets`);
+$$.chdir(`stories`);
+```
+
+Create layout sample
+
+```typescript
+$$(`mkdir -p layouts`);
+$$.chdir(`layouts`);
+```
+
+```typescript
+# |-> StandardLayout.tsx
+import React from 'react';
+
+export const StandardLayout: React.FC = ({ children }) => (
+  <div className="max-w-7xl mx-auto">
+    <header className="text-4xl p-4 bg-blue-50">Header</header>
+    <main className="p-4 prose">{children}</main>
+    <footer className="p-4 text-xs bg-blue-900 text-white">Footer</footer>
+  </div>
+);
+```
+
+```typescript
+# |-> StandardLayout.stories.tsx
+import { Meta, Story } from '@storybook/react';
+import React from 'react';
+
+import { StandardLayout } from './StandardLayout';
+
+export default {
+  title: 'Components/Layouts/Standard',
+  component: StandardLayout,
+  parameters: {
+    layout: 'fullscreen',
+  },
+} as Meta;
+
+export const Standard: Story = () => (
+  <StandardLayout>
+    <div className="border-2 border-gray-300 border-solid h-24" />
+  </StandardLayout>
+);
+```
--- a/packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md
+++ b/packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md
@@ -57,3 +57,15 @@ interfaces
 // add react & react-dom
 $$('yarn add react react-dom');
 ```
+
+### Storybook
+
+Add **Storybook**, an open source tool for developing UI components and pages in
+isolation. It simplifies building, documenting, and testing UIs.
+
+```typescript
+// add npx
+$$('yarn add npx');
+// install & initialize Storybook
+$$('npx sb init');
+```
--- a/packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md
+++ b/packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md
@@ -33,7 +33,7 @@ Add the project namespace
 // add the project namespace
 $$.file('package.json', (json) => ({
   ...json,
-  name: `${projectName}-ui`,
+  name: `@${projectName}/ui`,
   description: `Design system for ${projectName.toUpperCase()}`,
 }));
 ```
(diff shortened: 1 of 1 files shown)


packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md
```diff
@@ -0,0 +1,215 @@
...
+// Check node version.
+$$('node -v', {
+  stdout: $$.minimalVersion('12'),
+});
+
+// Check yarn version.
+$$('yarn -v', {
+  stdout: $$.minimalVersion('1.0'),
+});
+```
```
pmelab: In the other recipes i added this snippet to make sure we are inside a monorepo and also fetch the project name.
Then we also don't have  to prompt again later.

```suggestion
> This recipe has to be run from the root folder of a mono-repository created
> with `amazee-recipes create-monorepo`.

```typescript
$$('cat README.md', {
  stdout: /Executed `create-monorepo`/,
});

const { name: projectName } = $$.file('package.json');
```
```

File: packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md [hunk 2]
```diff
@@ -0,0 +1,215 @@
...
+// Choose a project name.
+const { projectName } = $$.prompts({
+  type: 'text',
+  name: 'projectName',
+  message: 'What is the name of the project?',
+  validate: (name) =>
+    !/^[a-z][a-z\d_]+$/.test(name)
+      ? 'Project names must start with a letter and contain lower case letters, numbers and underscores only.'
+      : true,
+});
```
pmelab: See above.

```suggestion
```

packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md
```diff
@@ -0,0 +1,215 @@
...
+
+$$(`mkdir -p packages/@${projectName}/ui`);
+$$.chdir(`packages/@${projectName}/ui`);
+```
+
+Initialize the Ui project folder with yarn
+
+```typescript
+// yarn init
+$$('yarn init -w -y');
```
pmelab: I think we have to add the project namespace, or else we can't reference the package properly with `import '@project/ui';`. And it has to be private, or lerna will try to publish it to npm (and fail because it can't publish to the `@project` namespace).

What does `-w` do?

```suggestion
$$('yarn init -w -y -p');
$$.file('package.json', (json) => ({
  ...json,
  name: `@${projectName}/ui`,
  description: `Design system for ${projectName.toUpperCase()}`,
});
```

packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md
```diff
@@ -0,0 +1,215 @@
...
+### Tailwind
+
+Add **Tailwind**, a utility-first CSS framework packed with classes like flex,
+pt-4, text-center and rotate-90 that can be composed to build any design,
+directly in your markup.
+
+```typescript
+// install Tailwind
+$$(
+  'yarn add tailwindcss@npm:@tailwindcss/postcss7-compat postcss@^7 autoprefixer@^9',
```
pmelab: Storybook does optionally support PostCSS 8: https://storybook.js.org/addons/@storybook/addon-postcss

I think we should rather add that configuration, instead of using the tailwind-compat package. I also ran into problems with other PostCSS extensions that require 8. We can simply remove it when Storybook ships with PostCSS 8 by default.

packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md
```diff
@@ -0,0 +1,215 @@
...
+```typescript
+$$(`mkdir -p layouts`);
+$$.chdir(`layouts`);
+```
+
+```typescript
+# |-> StandardLayout.tsx
+import React from 'react';
+
+export const StandardLayout: React.FC = ({ children }) => (
```
pmelab: @marcotisi has an [opinion on React.FC](https://github.com/gndelia/codemod-replace-react-fc-typescript#man_teacher-motivation) (or why not to use it) 😃 
```suggestion
export const StandardLayout = ({ children }: React.PropsWithChildren) => (
```

pmelab (changes requested):
Good work! I have a couple of suggestions (apart from the once commented inline):
* The recipe should make sure everything is committed at the end and make sure there are no leftovers, [like the other recipes do](https://github.com/AmazeeLabs/silverback-mono/blob/development/packages/npm/%40amazeelabs/recipes/recipes/add-gatsby.ts.md) 
* It should also run tests, to make sure it does not create something that's broken upfront.

packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md
marcotisi: Yeah, let's avoid `FC`, `VFC`, and `SFC`. If you need more details, ping me so I can explain 😉

packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md
marcotisi: Yeah, you don't need the `compat` version because the `ui` package is separate from the website and compiles itself down to `es5`.

packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md
```diff
@@ -0,0 +1,243 @@
...
+Create layout sample
+
+```typescript
+$$(`mkdir -p layouts`);
+$$.chdir(`layouts`);
+```
+
+```typescript
+# |-> StandardLayout.tsx
+import React, {PropsWithChildren} from 'react';
```
marcotisi:
```suggestion
import React, { PropsWithChildren } from 'react';
```

packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md
```diff
@@ -0,0 +1,243 @@
...
+$$('yarn init -y -p');
+```
+
+Add the project namespace
+
+```typescript
+// add the project namespace
+$$.file('package.json', (json) => ({
+  ...json,
+  name: `${projectName}-ui`,
```
marcotisi: Let's name the UI package `@{projectName}/ui`:

```suggestion
  name: `@${projectName}/ui`,
```

packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md
```diff
@@ -0,0 +1,243 @@
...
+```typescript
+// Pick project name from package.json.
+const { name: projectName } = $$.file('package.json');
+```
+
+Create a folders structure to contain the UI
+
+```typescript
+$$(`mkdir -p packages/${projectName}/ui`);
+$$.chdir(`packages/${projectName}/ui`);
```
marcotisi: Let's put the package inside `/packages/@{$projectName}/ui` folder:

```suggestion
$$(`mkdir -p packages/@${projectName}/ui`);
$$.chdir(`packages/@${projectName}/ui`);
```

packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md
```diff
@@ -0,0 +1,243 @@
...
+@tailwind components;
+@tailwind utilities;
+```
+
+Update tailwind.config.js
+
+```typescript
+# |-> tailwind.config.js
+module.exports = {
+  purge: ['./src/**/*.tsx'],
```
marcotisi, line 129: Let's enable Tailwind's just in time compiler:

```suggestion
  mode: 'jit',
  purge: ['./src/**/*.tsx'],
```

marcotisi (changes requested):
Still some minor adjustment and we're ready to go

packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md
mattiasimonato: I've intentionally used '-' instead of '/' otherwise it will break recipes build cause the name is not valid for the regex rule.
Testing it inside '@amazeelabs/recipes' so the result would be '@amazeelabs/recipes/ui' which is not valid

packages/npm/@amazeelabs/recipes/recipes/add-storybook.ts.md
mattiasimonato: same as above

marcotisi (approved):

marcotisi (changes requested):
Let's fix all the issues we found in [this commit](https://github.com/AmazeeLabs/gatsby-workshop/commit/11e78eace1ade407441da7392492922bd8808479) and refactor the folder structure using [this pattern](https://github.com/AmazeeLabs/silverback-mono/issues/450#issuecomment-787836763).

pmelab (approved):

PR #671, opened 2021-06-18
Simplify translation handling

pmelab:
## Package(s) involved

* `silverback_gatsby`
* `gatsby-source-silverback`

## Description of changes

Simplify the multilingual GraphQL schema by removing the need for the "wrapper" type '`*Translations`. Every translatable type is the full object with additional translation properties.

* `translations`: the list of versions in other languages
* `defaultTranslation`: boolean indicating if this object is the default translation
* `langcode`: the language id of this object

Every translation is now treated as a solitary node in Gatsby. To achieve this, the `id` property on translatable types now also includes the langcode (`123:de`) and a custom pagination adapter *flattens* translations when sourcing them into Gatsby. That way we can use standard entity query (which can't return rows for each language) but achieve a simpler API in Gatsby level.

*All* types (translatable and language agnostic ones) now contain two ID properties:

* `id`: the id used by gatsby that is unique among translations of the same item
* `drupalId`: drupal internal id that is shared across all translations

With these properties it is rather easy to query the Gatsby data store for items in a specific language or search for translations of a specific item.

```graphql
query AllGermanPages {
  allDrupalPage(filter: {langcode: {eq: "de"}}) {
    title
  }
}

query GermanTranslation($drupalId: String!) {
  drupalPage(filter: {langcode: {eq: "de"}, drupalId: {eq: $drupalId }}) {
    title
  }
}
```

The `translations` property returns a list of all translations of the item, while `translation` allows to search for a specific language. The latter returns the default translation if the requested language does not exist.

## Motivation and context

Before this, connecting a type to a translatable data source, would spawn an additional "wrapper" type with suffix `*Translations` that is managed by the system entirely, so developers don't have to do this manually all the time.

```graphql
type Article @entity(type: "node", bundle: "article") {
  title
  # Auto-created
  langcode: String!
}

# Auto-created
type ArticleTranslations {
  id: String!
  translations: [Article!]!
}
```

This resulted in two types of DX problems:

On the one hand, the query fields are named inconsistently. A translatable type would generate Gatsby root level fields `allDrupal[Type]Translations` and `drupal[Type]Translations`, while un-translatable ones are called `allDrupal[Type]` and `drupal[Type]`. It's hard to explain and remember which applies when and switching between translated and un-translated breaks a lot. Now simply new properties appear on the same type.

The bigger issue was referencing translatable types. It required to use a type that does not exist, which breaks some tooling and feels to magical.

```graphql
# Media type image is translatable in this case
type Image @entity(type: "media", bundle: "image") {
  url: String!
  alt: String!
}

type Article @entity(type: "node", bundle: "article") {
  title: String!
  # We have to use the wrapper here, because "Image" is
  # not a Gatsby node type that can be referenced.
  image: ImageTranslations
}
```

Now we can just use the types declared in the schema.

```graphql
# Media type image is translatable in this case
type Image @entity(type: "media", bundle: "image") {
  url: String!
  alt: String!
}

type Article @entity(type: "node", bundle: "article") {
  title: String!
  image: Image
}
```

And query translations of referenced items just within the query:

```graphql
query ArticlesWithImages($langcode: String!) {
  allDrupalArticle(filter: {langcode: {eq: $langcode}}) {
    title
    image {
      url
      translation(langcode: $langcode) {
        alt
      }
    }
  }
}
```

## How has this been tested?
The unit tests for the `silverback_gatsby` module have been extended. Additionally the `Image` media type in the `silverback-gatsby` example application has been changed to "translatable" and the queries have been adjusted, while the existing tests still pass.


packages/composer/amazeelabs/silverback_gatsby/src/GatsbyUpdateTrigger.php
```diff
@@ -53,9 +53,10 @@ public function sendUpdates() : void {
             'json' => ['buildId' => $id],
           ]);
           if (
-            \Drupal::moduleHandler()->moduleExists('gatsby_build_monitor') &&
+            false
+//            \Drupal::moduleHandler()->moduleExists('gatsby_build_monitor') &&
```
Leksat: Was it breaking something?

Leksat (commented):
Looks awesome in general!

packages/composer/amazeelabs/silverback_gatsby/src/GatsbyUpdateTrigger.php
pmelab: Oh, damn. Forgot that. Yes, it broke the unit tests.

Leksat: maybe if it's replaced with `function_exists('_gatsby_build_monitor_state')`, it won't affect unit tests?

pmelab: Just did that 😄

chindris (approved):

PR #679, opened 2021-06-24
feat: avoid ajax call when tab is not active

rodrigoaguilera:
## Package(s) involved
drupal/gatsby_build_monitor

## Description of changes
Only refresh the status on the toolbar when the browser tab is active

## Motivation and context
Reduce the load on requests made to the Drupal backend

## How has this been tested?
Just locally


Leksat:
@rodrigoaguilera it was a glitch. I re-ran it and it's green 🤷

Issue #688, opened 2021-07-01
Gatsby: Handle access gracefully


pmelab:
Currently the gatsby list queries break when they encounter an entity the user is not allowed to access. This is often a problem with more complex permission handlers (e.g. commerce product variations which depend on their product). In these cases the list should rather just emit an empty object that can be filtered in the custom pagination handler.

PR #689, opened 2021-07-01
feat(gatsby): nullify inaccessible entities instead of failing

pmelab:
## Package(s) involved
* `silverback_gatsby`

## Description of changes
Instead of failing completely, the `list_entities` data producer simply returns a `null` item if an entity is not accessible by the current user.

## Motivation and context
In case of more complex permission models, the build completely failed if the entity query returned an entity that is not accessible. Now it won't fail, but just emit an empty item that is ignored by the source plugin.

## Related Issue(s)
#688 

## How has this been tested?
With a unit test and the integration tests in `silverback-gatsby` should still pass.

### DIFF
--- a/packages/composer/amazeelabs/silverback_gatsby/src/Plugin/GraphQL/DataProducer/ListEntities.php
+++ b/packages/composer/amazeelabs/silverback_gatsby/src/Plugin/GraphQL/DataProducer/ListEntities.php
@@ -3,6 +3,7 @@
 namespace Drupal\silverback_gatsby\Plugin\GraphQL\DataProducer;
 
 use Drupal\Core\Cache\RefinableCacheableDependencyInterface;
+use Drupal\Core\Entity\EntityInterface;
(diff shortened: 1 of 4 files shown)


packages/composer/amazeelabs/silverback_gatsby/src/Plugin/GraphQL/DataProducer/ListEntities.php
```diff
@@ -58,11 +59,12 @@ public function resolve(string $type, ?string $bundle, ?int $offset, ?int $limit
       $metadata->addCacheableDependency($entity);
     }
 
-    if ($access) {
-      $this->checkAccess($entities);
```
Leksat, line 62: AFAIK this method is not used anywhere else so we can delete it from `EntityQueryBase`.

Leksat (approved):
Sounds reasonable!

PR #690, opened 2021-07-02
Proxy packages

pmelab:
## Description of changes
Create a proxy-package for `drupal/gutenberg` that applies necessary patches.

## Motivation and context
Alternative solution for #644 , that does not require a dedicated process. Patches are managed in proxy-packages and projects can simply install them instead of the packages themselves.


apps/silverback-drupal/composer.lock
```diff
@@ -75,6 +75,36 @@
             },
             "time": "2021-04-13T14:13:44+00:00"
         },
+        {
+            "name": "amazeelabs/proxy-gutenberg",
+            "version": "1.0.0",
+            "dist": {
+                "type": "path",
+                "url": "../../packages/composer/amazeelabs/proxy-gutenberg",
```
fjgarlin, line 83: This reference is correct within this context, but on a separate project it will not. I assume that we'd just need to add a `repository` section into the `composer.json` file for this, right?

Leksat (approved):
Let's merge it and see if it works.

apps/silverback-drupal/composer.lock
pmelab: This is only here. In another project it would just be loaded from packagist.

fjgarlin: Perfect.

Leksat:
Tested. Works!

Just a reminder. Patching will work only if the root `composer.json`
- has `cweagans/composer-patches` in dependencies
- has `"enable-patching": true` in `extra` settings

Issue #717, opened 2021-07-12
Gatsby Preview toolbar button


pmelab:
Integrate a button into the Drupal toolbar that opens the Gatsby preview for the currently displayed page in a new window. Optionally with a screen size selection to quickly test content on mobile/tablet/desktop.

PR #725, opened 2021-07-12
feat(slb-144): initial work on toolbar external preview button

nbaosullivan:
## Package(s) involved
`@amazeelabs/silverback_external_preview`

## Description of changes
Provides a UI to preview pages (published and draft) on an external website - usually a Gatsby or NextJS frontend.

## Motivation and context
When using a decoupled architecture, there needs to be an easy way to preview content on the frontend.

## Related Issue(s)
#717


pmelab (changes requested):
Awesome work! 🥳 

The only thing I would change right now is the coloring. The low contrast makes the device icons hard to read and the main preview button would look friendlier in some color.
Apart from that i think we are good to merge the first version!

nbaosullivan:
> Awesome work! 🥳
> 
> The only thing I would change right now is the coloring. The low contrast makes the device icons hard to read and the main preview button would look friendlier in some color.
> Apart from that i think we are good to merge the first version!

Some styling changes have been made!

PR #739, opened 2021-07-21
fix(gatsby): properly handle null in gatsby result lists

pmelab:
Was introduced with: https://github.com/AmazeeLabs/silverback-mono/pull/689

## Package(s) involved
`gatsby-source-silverback`

## Description of changes
#689 introduced possible null entries in Gatsby result lists, but this part of the plugin was not prepared for that yet.


chindris:
I just did these changes locally and the null issue was fixed, so this looks good from my side.

PR #746, opened 2021-07-26
fix: do not process urls with schemes

chindris:
## Package(s) involved
`silverback_gutenberg`

## Description of changes
When processing the urls from the Gutenberg editor, only specific schemes (currently http and https) are considered. Other urls, like mailto, telephone, etc, should stay exactly how they are.

## How has this been tested?
Manual, locally.


Leksat (changes requested):
I would update `LinkProcessor::isExternal` to return `true` for `mailto` and similar links.

And then we could add a test case to `LinkProcessorTest::testProcessUrl`.

Leksat (approved):

PR #751, opened 2021-07-26
chore(storybook eslint tailwind): issue/713 Add eslint-plugin-tailwindcss

stewest:
chore(storybook eslint tailwind): issue/713 Add eslint-plugin-tailwindcss
chore(storybook eslint tailwind): issue/713 Adds recipse for eslintrc config


pmelab:
I fear this is not going to work ☹️ Everything that's created by `@amazeelabs/scaffold` is re-created every time somebody updates the project. The recipe runs only once. So the changes you make in the recipe won't last. The idea behind that is that we can control these configuration files across multiple projects.

We would have to add that to the [central `eslint`](https://github.com/AmazeeLabs/silverback-mono/blob/development/packages/npm/%40amazeelabs/eslint-config/index.js) config that is loaded by the scaffolded one. That way we add the tailwind linter to all projects retroactively.

stewest:
change to this suggestion from PM:

I fear this is not going to work ☹️ Everything that's created by @amazeelabs/scaffold is re-created every time somebody updates the project. The recipe runs only once. So the changes you make in the recipe won't last. The idea behind that is that we can control these configuration files across multiple projects.

We would have to add that to the central eslint config that is loaded by the scaffolded one. That way we add the tailwind linter to all projects retroactively.

https://github.com/AmazeeLabs/silverback-mono/blob/development/packages/npm/%40amazeelabs/eslint-config/index.js

stewest:
Hmm, some drupal build test failure.

pmelab:
Yeah, we still have some random failures. But seems to work now. Thanks @stewest !

PR #754, opened 2021-07-28
feat(gatsby): create gatsby pages automatically

Leksat:
## Package(s) involved

- `composer/amazeelabs/silverback_gatsby`
- `npm/@amazeelabs/gatsby-source-silverback`

## Description of changes

- Added `@path` and `@template` directives for automatic Gatsby pages creation. Docs: https://github.com/AmazeeLabs/silverback-mono/blob/create-gatsby-pages-automatically/packages/npm/@amazeelabs/gatsby-source-silverback/README.md#automatic-creation-of-gatsby-pages
- Some cleanup.

## Related Issue(s)

https://github.com/AmazeeLabs/silverback-mono/issues/707

## How has this been tested?

- Extended `composer/amazeelabs/silverback_gatsby` unit tests.
- Extended `apps/silverback-gatsby` integration tests to check `@template` cases:
  - stub template for missing template file
  - default template name
  - custom template name


apps/silverback-gatsby/src/templates/article.tsx
```diff
@@ -21,11 +22,9 @@ export const query = graphql`
...
       }
       image {
-        translation(langcode: $langcode) {
-          alt
-          localImage {
-            ...ImageSharpFixed
-          }
+        alt
+        localImage {
+          ...ImageSharpFixed
```
pmelab, line 27: Do we still get the `alt` text in the correct language?

Leksat: I'm not sure. Maybe. But if this is the case, it should be fixed in https://github.com/drupal-graphql/graphql/pull/974

Vasi already uses this patch on his project.

Leksat: Actually, even if it is fixed in the upstream, there still can be issues. E.g.
- a translatable entity references an untranslatable one, which in turn references a translatable entity
  - the language will be lost on the interim entity
- an untranslatable page references a translatable entity
  - here we don't know the current locale in the page template

I'd say, let's just see how it goes 😅 In most cases all entities are translatable and it's a rare case when we have more than one reference level.

pmelab: Its not just about entity references. What about entities that are embedded in gutenberg documents?

pmelab: I think `load[Type]` should set the current language context and all sub-fields should respect that when loading other entities. But that is not in scope of this PR.

PR #771, opened 2021-08-11
feat(cdn-redirect): add missing translation redirect

Leksat:
## Package(s) involved

- `composer/amazeelabs/silverback_cdn_redirect`

## Description of changes

Added a redirect to the source entity language if a non existing entity translation was requested.

It is disabled by default. Can be enabled like this:
```
drush cset silverback_cdn_redirect.settings missing_translation_redirect_entity_types '["node", "commerce_product"]' --input-format=yaml
```

## How has this been tested?

Locally on a project.

### DIFF
--- a/packages/composer/amazeelabs/silverback_cdn_redirect/src/EventSubscriber/MissingTranslationRedirectSubscriber.php
+++ b/packages/composer/amazeelabs/silverback_cdn_redirect/src/EventSubscriber/MissingTranslationRedirectSubscriber.php
@@ -1,0 +1,90 @@
+<?php
+
+namespace Drupal\silverback_cdn_redirect\EventSubscriber;
+
+use Drupal\Core\Config\ConfigFactoryInterface;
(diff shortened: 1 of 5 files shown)


packages/composer/amazeelabs/silverback_cdn_redirect/src/EventSubscriber/MissingTranslationRedirectSubscriber.php
```diff
@@ -0,0 +1,90 @@
...
+          ] + $event->getRequest()->query->all(),
+        ]
+      );
+      $response = new TrustedRedirectResponse($url->toString());
+
+      // Add the necessary cache contexts to the response, as redirect
+      // responses are cached as well.
+      $response->getCacheableMetadata()
+        ->addCacheContexts([
+          'languages:language_content',
```
Leksat, line 81: @chindris this is the only change from your original code. It was `languages:language_interface` here.

pmelab (approved):

PR #765, opened 2021-08-06
Variable existence check is introduced in Conditions class to avoid notice errors

maijs:
## Package(s) involved
Package `drupal/webform_jsonschema` is affected.

## Description of changes
`Conditions::prepareDependencySchema()` method contains `if` statements where existence of variables is not checked. This causes PHP notice errors to be issued and automated tests fail due to non-0 code returns.

## Related Issue(s)
Original report is posted on drupal.org - https://www.drupal.org/project/webform_jsonschema/issues/3225332

## How has this been tested?
This has been manually tested by running the tests and not experiencing notice errors.

PR #780, opened 2021-08-16
Coalescing operator expression is wrapped in parentheses in Condition class

maijs:
This is a follow-up PR to https://github.com/AmazeeLabs/silverback-mono/pull/765.

## Package(s) involved
Package `drupal/webform_jsonschema` is affected.

## Description of changes
The coalescing operator `??` needs to be wrapped in parentheses in order to work with comparison. Here's the code to check the behaviour:

```php
<?php

$schema['properties']['name']['type'] = 'string';

if ($schema['properties']['name']['type'] ?? NULL === 'boolean') {
  print_r('Element type is not a boolean.');
}

if (($schema['properties']['name']['type'] ?? NULL) === 'boolean') {
  print_r('This should never print.');
}
```

The proposed resolution involves wrapping coalescing operator expressing in parentheses so that comparison yields correct results:

```php
if (($schema['properties'][$dependencyKey]['type'] ?? NULL) === 'boolean' && $trigger === 'filled') {
// ...
}

if (($schema['properties'][$dependencyKey]['type'] ?? NULL) === 'boolean') {
// ...
}
```

## Related Issue(s)
Original report is posted on drupal.org - https://www.drupal.org/project/webform_jsonschema/issues/3225332

## How has this been tested?
This change has been manually tested.

### DIFF
--- a/packages/composer/drupal/webform_jsonschema/src/Conditions.php
+++ b/packages/composer/drupal/webform_jsonschema/src/Conditions.php
@@ -117,7 +117,7 @@ class Conditions {
     $value = reset($triggerArray);
     $trigger = key($triggerArray);
 
-    if ($schema['properties'][$dependencyKey]['type'] ?? NULL === 'boolean' && $trigger === 'filled') {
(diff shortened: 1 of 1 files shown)


Leksat:
Thanks! Released in 2.0.2

Issue #718, opened 2021-07-12
Drupal autosave & live update support


pmelab:
We need auto-save support for Drupal pages.

- [ ] Works with Gutenberg
- [ ] Integrate with `silverback_gatsby` to show auto-save versions in preview

nbaosullivan:
Notes:
- Look at extending https://www.drupal.org/project/autosave_form to work with Gutenberg
- Think about required fields https://www.drupal.org/project/require_on_publish
- Read the discussion here: https://www.drupal.org/project/drupal/issues/3029488

PR #781, opened 2021-08-16
feat(slb-146): include autosave_form and patch to provide support for…

nbaosullivan:
## Package(s) involved
`amazeelabs/proxy-gutenberg`
`amazeelabs/proxy-autosave-form`

## Description of changes
Include Autosave Form module to Drupal and patch to support Gutenberg

## Related Issue(s)
#718 
[[SLB-146] Autosave for Gutenberg and Gatsby](https://jira.amazee.com/browse/SLB-146)

## How has this been tested?
Locally


pmelab (changes requested):
Great work so far! 🥳 

I wonder if we actually need the autosave patch? Couldn't we just write gutenberg content periodically? We know that this might stay a patch for a long time 🙄 And it wouldn't create a dependency between the two.

**Next steps:**
* extend the [Fetch Entity](https://github.com/AmazeeLabs/silverback-mono/blob/development/packages/composer/amazeelabs/silverback_gatsby/src/Plugin/GraphQL/DataProducer/FetchEntity.php) data producer to return the latest autosaved values when the user is allowed to see unpublished content
* [trigger an update](https://github.com/AmazeeLabs/silverback-mono/blob/development/packages/composer/amazeelabs/silverback_gatsby/silverback_gatsby.module#L9) a new autosave is stored

PR #793, opened 2021-09-01
Silverback Iframe & Playwright

Leksat:
## Description of changes

Moved all our iframe related work to Silverback.

I split the work into several commits. Maybe it can help review the PR.

## Motivation and context

To not have duplicated code in projects.

## How has this been tested?

Mainly with Playwright tests.

### What can be improved in the test runner

Currently Drupal authentication is solved with `drupalLogin`/`drupalLogout` helpers. This works, but it should be possible to re-do the auth with https://playwright.dev/docs/test-fixtures/. The idea is to use fixtures to extend `test`. In addition to `page` argument, it could provide `authPage` which could be used to make changes via Drupal UI.

I tried to implement the above, but it did not work. Playwright were just hanging. But maybe I did something wrong.

Also, I tried to use https://playwright.dev/docs/auth/#reuse-authentication-in-playwright-test. It worked but did not make a lot of sense because in the tests I needed both authorized and anonymous contexts.

## Other findings

[zx](https://www.npmjs.com/package/zx) (+ [tcp-port-used](https://www.npmjs.com/package/tcp-port-used) + [cross-port-killer](https://www.npmjs.com/package/cross-port-killer)) - I will never write a bash script again 😄


packages/composer/amazeelabs/silverback_iframe_theme/silverback_iframe_theme.info.yml
```diff
@@ -0,0 +1,37 @@
...
+      theme:
+        css/commerce_checkout.progress.css: false
+  commerce_checkout/form:
+    css:
+      theme:
+        css/commerce_checkout.form.css: false
+  address/form:
+    css:
+      theme:
+        css/address.form.css: false
```
pmelab: Those excludes feel arbitrary. I think we should either remove the entire CSS layer (probably `theme` and `layout`?) entirely, or we leave these untouched and leave it to the projects sub-theme to exclude unused styles. In cases where the embedded pages are not central to the actual product, we might actually want these styles and just slightly adjust them.

apps/silverback-drupal/web/modules/custom/silverback_gatsby_test/src/Entity/Server.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace Drupal\silverback_gatsby_test\Entity;
+
+use Drupal\Core\Utility\Error;
+use Drupal\graphql\Entity\Server as ServerOriginal;
+use Drupal\graphql\GraphQL\Execution\ExecutionResult as CacheableExecutionResult;
+use GraphQL\Server\OperationParams;
+
+// TODO: Remove once https://github.com/drupal-graphql/graphql/issues/1195 is
```
pmelab, line 10: Shouldn't we create a patch and a proxy package instead?

apps/silverback-drupal/web/modules/custom/silverback_gatsby_test/src/Plugin/GraphQL/Schema/SilverbackGatsbyTestSchema.php
```diff
@@ -93,6 +93,19 @@ public function getResolverRegistry(): ResolverRegistryInterface {
       $builder->produce('url_path')->map('url', $builder->fromParent()),
     ));
 
+    $addResolver('Webform.path',
+      $builder->compose(
+        $builder->produce('entity_url')->map('entity', $builder->fromParent()),
+        $builder->produce('url_path')->map('url', $builder->fromParent())
+      )
```
pmelab, line 100: This is repeated _a lot_. Wouldn't it make sense to also auto-create these resolvers when we have a `@path` directive on an `@entity` type?

apps/silverback-drupal/web/modules/custom/silverback_gatsby_test/src/Plugin/GraphQL/Schema/SilverbackGatsbyTestSchema.php
```diff
@@ -93,6 +93,19 @@ public function getResolverRegistry(): ResolverRegistryInterface {
...
+    );
+    $addResolver('Webform.url',
+      $builder->compose(
+        $builder->produce('entity_url')->map('entity', $builder->fromParent()),
+        $builder->callback(fn (Url $url) => $url->setAbsolute()->toString(TRUE)->getGeneratedUrl())
+      )
```
pmelab, line 106: Also a candidate for a dedicated `@url` directive with auto-resolvers.

packages/npm/@amazeelabs/silverback-playwright/src/drupal-auth.ts
```diff
@@ -0,0 +1,13 @@
+import { PlaywrightTestArgs } from '@playwright/test';
+
+import { drupal } from './constants';
+
+export const drupalLogin = async (page: PlaywrightTestArgs['page']) => {
+  await page.goto(
+    `${drupal.baseUrl}/test-session/set?X-TEST-SESSION-USER=${drupal.adminUser.login}`,
+  );
```
pmelab, line 8: We should try to get rid of `test-session` and use playwrights `auth` capabilities instead, since we won't have to fiddle with Drupal internals and potentially create situations where it behaves differently when under test.

pmelab (approved):
Brilliant work! 😲 

I added a couple of improvement suggestions, but I think they could just be turned into follow-up issues.

packages/npm/@amazeelabs/silverback-playwright/src/drupal-auth.ts
pmelab: > Also, I tried to use https://playwright.dev/docs/auth/#reuse-authentication-in-playwright-test. It worked but did not make a lot of sense because in the tests I needed both authorized and anonymous contexts.

Wouldn't we achieve that with different page contexts?

packages/composer/amazeelabs/silverback_iframe_theme/silverback_iframe_theme.info.yml
Leksat: I just copied this part over 🙈 

Project's sub-theme sounds great. Because we still need to add project-specific CSS.

I'll remove the overrides in this PR to not have breaking changes later.

Leksat: But, wait, how do we know them which theme to enable when `iframe=true` is in URL?

apps/silverback-drupal/web/modules/custom/silverback_gatsby_test/src/Entity/Server.php
Leksat: Yeah, it's my plan to submit a proper patch to https://github.com/drupal-graphql/graphql. Just need time 😬

apps/silverback-drupal/web/modules/custom/silverback_gatsby_test/src/Plugin/GraphQL/Schema/SilverbackGatsbyTestSchema.php
Leksat: Yeah, `@isPath` and `@producePath` IIRC what we have discussed.

apps/silverback-drupal/web/modules/custom/silverback_gatsby_test/src/Plugin/GraphQL/Schema/SilverbackGatsbyTestSchema.php
Leksat: `@produceUrl` 😉

packages/npm/@amazeelabs/silverback-playwright/src/drupal-auth.ts
Leksat: When I tried the "reuse auth" feature, it worked very strange. I tried to create a new context+page in a test.

The documentation of [`browser.newContext()`](https://playwright.dev/docs/api/class-browser#browser-new-context) says:

> Creates a new browser context. It won't share cookies/cache with other browser contexts.

But for me the new page still had Drupal session cookies 🤷 

Maybe we could try together at some point?

packages/composer/amazeelabs/silverback_iframe_theme/silverback_iframe_theme.info.yml
Leksat: Addressed in https://github.com/AmazeeLabs/silverback-mono/pull/793/commits/970d5e9ed02ea2d8537c83fe81ae062f6062e86f

PR #802, opened 2021-09-08
Feature/accessiblity tests integration

luqmaanessop:


luqmaanessop:
This will need some more thinking, as this works OK for the page with TOC but because I am using a layout decorator the a11y checks skips the header and footer by default and is only checking the inside page

Issue #787, opened 2021-08-23
IFrame theme and webforms


Leksat:
We did some work on a client project:
- created iframe theme to display Drupal forms on Gutenberg pages (in iframe)
- implemented several webform confirmation options (redirect to url, display message, etc.)

Now we want to move this to Silverback and cover it with integration tests.

Leksat:
Done in https://github.com/AmazeeLabs/silverback-mono/pull/793 + https://github.com/AmazeeLabs/silverback-mono/commit/d598d953b4e8651456cea75891c97e988e485028

Leksat:
https://github.com/AmazeeLabs/silverback-mono/blob/3f0046c611465996e36b243afb4e0821c9b1e82f/packages/composer/amazeelabs/silverback_iframe_theme/iframeCommand.js#L95-L111
We need to exclude Drupal AJAX links here.

PR #803, opened 2021-09-09
silverback_iframe: exclude some links from processing

Leksat:
## Package(s) involved

- `composer/amazeelabs/silverback_iframe_theme`

## Motivation and context

The link processing code broke Commerce Checkout links.

Commerce links are not AJAX, so I had to exclude links based on classes.

## Related Issue(s)

https://github.com/AmazeeLabs/silverback-mono/issues/787

## How has this been tested?

Locally on a project. Automated testing would need Commerce which I don't want to add to `silverback-drupal`.


pmelab (approved):
That's good for now. Perhaps we should provide config or a hook in the future, so projects can adjust these.

PR #811, opened 2021-09-14
Cypress -> Playwright

### LINKED ISSUE #809: Cypress -> Playwright
Cypress tests from `silverback-gatsby` app should be rewritten with Playwright and moved to relevant packages.

Leksat:
## Description of changes

Cypress tests from `silverback-gatsby` app were rewritten with Playwright and moved to relevant packages.

See commit messages for additional smaller changes.

## Playwright experience

There were few fixes/improvements for `@amazeelabs/silverback-playwright`. Almost no issues with the Playwright itself.

Some notes:

1. Playwright might be too fast. E.g. it can click stuff before JS behaviors are attached to elements. We can try different `waitUntil` options of [`page.goto`](https://playwright.dev/docs/api/class-page#page-goto) if this becomes a problem.
2. Same thing with Gatsby. Playwright can start interact with a Gatsby page before Gatsby is fully initialized. As we discussed, we maybe can use Gatsby's [built-in Cypress support](https://github.com/gatsbyjs/gatsby/blob/4a765b5c62208d58f0bd7fd59558160c0b9feed3/packages/gatsby/cache-dir/api-runner-browser.js#L7-L15) to deal with this.
3. There is https://www.npmjs.com/package/playwright-testing-library. But Playwright has nice built-in ways to do same things. For example, we can address input elements by their labels:
    ```ts
    await page.uncheck('label:text-is("Published")');
    ```
    Also, there is a variety of [text selectors](https://playwright.dev/docs/selectors/#text-selector) to use.
4. Found a glitch with contenteditable elements. Using [`page.fill`](https://playwright.dev/docs/api/class-page#page-fill) works perfectly with standard inputs. However, when applied to contenteditable, it can sometimes append the text instead of replacing it. Yet this can be solved with `page.fill(''); page.fill('my text')`.
5. When googling stuff, I sometimes type "cypress" instead of "playwright" 😅 

## Related Issue(s)

Closes https://github.com/AmazeeLabs/silverback-mono/issues/809

## How has this been tested?

Playwright!


pmelab (approved):
🥳 🎆 ✅

PR #835, opened 2021-10-11
Recipe test workflow

pmelab:
## Package(s) involved
* `@amazeelabs/recipes`
* `@amazeelabs/scaffold`

## Description of changes
The standard test suite does not run the full recipe and scaffold tests any more. Instead all recipes are tested once a week (which also covers scaffold, since it's used there).

## Motivation and context
Recipes and scaffolding don't adhere to any version locking. They are meant to be "executable tutorials" that just rely on the latest version of employed packages. This means that they can unexpectedly break on any pull request and - on the other hand - they can break unnoticed if we don't have package changes that trigger their tests.


Leksat (approved):

Leksat:
Maybe let's also add `workflow_dispatch`? To be able to run it manually.

PR #846, opened 2021-10-18
fix: restrict preview link to content entity types only

Leksat:
## Package(s) involved

- `composer/amazeelabs/silverback_external_preview`

## Description of changes

Fixed a fatal error on field edit routes.

## Motivation and context

We had the following error when trying to edit a field config
(e.g. `/admin/structure/types/manage/foo/fields/node.foo.field_bar`)
```
Drupal\Core\Entity\Exception\UndefinedLinkTemplateException: No link template 'edit-form' found for the 'field_config' entity type
```

## How has this been tested?

Locally. On a project.


nbaosullivan (approved):
All looks good. Thanks for the quick fix

PR #870, opened 2021-11-28
Gatsby: Support internal DNS when sourcing data

pmelab:
### Current state
Absolute URL’s in the GraphQL response will inherit the request host by default. When using internal routing (e.g. `http://nginx:8080`) for fetching data, this host will be emitted when Drupal generates absolute Url’s to itself (e.g. image style urls).

### Solution
Allow to set an “external” drupal url in the configuration of `gatsby-source-silverback` that will be used to send `X-Forwarded-*` headers and trigger Drupal’s reverse proxy behavior (which the build script technically is).

—

Created via [Raycast](https://www.raycast.com?ref=signatureGithub)


Leksat (approved):
Looks like it should work :)

PR #880, opened 2021-12-04
chore(docs): updates links

stewest:
It looks like in this instance the module is called `modules/custom/silverback_gatsby_test` rather than `modules/custom/silverback_gatsby`. So if this is the case, then merge, else disregard.


Leksat:
Thanks @stewest!

Issue #882, opened 2021-12-10
Published translations of untranslated entites are not listed


pmelab:
The `list_entities` resolver does not retrieve published translations where the source language is not published. `gatsby-source-silverback` expects them to be there and breaks in this scenario.

Listings should fall back to the first translation if the source is not published.

—

Created via [Raycast](https://www.raycast.com?ref=signatureGithub)

Issue #849, opened 2021-10-21
Default limit and offset for list_entities resolver


pmelab:
The Gatsby resolver for listing entities currently forces us to add `limit` and `offset` which is annoying while testing the API. There should be default values.

PR #883, opened 2021-12-10
Gatsby list published translations & non-required limit and offset

pmelab:
## Package(s) involved

## Description of changes
* Make listing queries fall back to the first published translation of an entity when the source language is not accessible.
* allow to omit `limit` and `offset` when running listing queries for debugging

## Related Issue(s)
#849 , #882 

## How has this been tested?
* unit test cases


Leksat:
Looks good!

PR #893, opened 2021-12-23
chore: update some drupal dependencies

Leksat:


Leksat:
Ah! Renovate does not create a PR in case if both are true:
- one of the new releases is younger than 2 hours
- update command fails

https://github.com/renovatebot/renovate/blob/347d7dc51fe18da8d25d96954d0995a6f1f00b7e/lib/workers/branch/index.ts#L419

So let's wait a bit and it will create a PR with the updated `composer.json` file.

PR #903, opened 2022-01-03
silverback_gatsby: fix schema export

Leksat:
Even the smartest PhpStorm inspections were not able to catch this. Because it's a PHP feature 😅


pmelab:
I don't get whats the difference. Where there definitions before?

Leksat:
Previous code was gluing definitions from different servers:
```php
foreach ($servers as $server) {
  //...
  $definition[] = $schema->getSchemaDefinition();
  //...
}
```

PR #936, opened 2022-01-31
Feature/a11y playfunctions integration tests

luqmaanessop:
## Motivation and context
Create failing play function to see if git blocks PR merge.
So that we may consider a11y issues more.


pmelab:
By now, this is covered by the new storybook test runner: https://storybook.js.org/blog/automate-accessibility-tests-with-storybook/

Issue #926, opened 2022-01-25
CDN redirect should redirect to aliases


pmelab:
When accessing a node via it's id on CDN redirect, the redirect should go to the alias if it exists, since only this path is rendered in the frontend.

—

Created via [Raycast](https://www.raycast.com?ref=signatureGithub)

chindris:
Hi @pmelab !
I have issues reproducing this issue... I tried it locally within the silverback drupal as well as on a running project (nuklearforum) and accessing a node id via the cdn redirect returns a proper response.
Example: curl -I https://cms.nuklearforum.ch/cdn-redirect/de/node/59

Theoretically, I could see it happen if, during a cdn redirect request, both redirect request subscribers would have to kick in (basially, if a cdn redirect request would need two redirects to get to the proper path). In that case, the CDN Redirect controller would basically stop after the first redirect. A possible fix would be to keep doing a request while the event which generated the response has stopped propagating (because that is an indication that there are possibly other subscribers which did not run and which could generate another redirect).

But in this specific case (accessing a node via its id), the RedirectRequestSubscriber should already respond with the path alias...

chindris:
Hi @pmelab!

without the redirect module, the issue can be reproduced. I added a fix for this in https://github.com/AmazeeLabs/silverback-mono/pull/944

PR #944, opened 2022-02-06
chore: redirect to the path alias of a route, if it exists

chindris:
## Package(s) involved

amazeelabs/silverback_cdn_redirect

## Description of changes

Redirect to the path alias of the current route, if it exists.

## Motivation and context

Accessing a cdn redirect route with, for example, the node id instead of the path alias would not perform any redirects. In the frontend, the only routes which are generated are the ones with aliases (if they have aliases) for a node.

## Related Issue(s)

[CDN redirect should redirect to aliases](https://github.com/AmazeeLabs/silverback-mono/issues/926)

## How has this been tested?

Manually, local.

### DIFF
--- a/packages/composer/amazeelabs/silverback_cdn_redirect/src/Controller/CdnRedirectController.php
+++ b/packages/composer/amazeelabs/silverback_cdn_redirect/src/Controller/CdnRedirectController.php
@@ -4,13 +4,37 @@ namespace Drupal\silverback_cdn_redirect\Controller;
 
 use Drupal\Core\Controller\ControllerBase;
 use Drupal\Core\Routing\TrustedRedirectResponse;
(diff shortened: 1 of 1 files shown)


Leksat:
@pmelab the newly added code only works if there is no Redirect module installed on the project.

Currently all projects have Redirect module enabled and redirects for `/node/123` work just fine. Because `Enforce clean and canonical URLs` option (`/admin/config/search/redirect/settings`) is enabled by efault.

Do we plan to have projects not having Redirect module?

pmelab:
We probably always have the redirect module. But currently CDN-redirect is broken without it, since Drupal might emit non-aliased links that wont exist in the frontend. And that results in a rewrite-loop. Also it breaks if the module is there but the option is turned off. So i think its good to have this as a safeguard?

pmelab (approved):

Issue #959, opened 2022-02-21
Multiple url encoding in internal links by the gutenberg url processor


chindris:
When a link is created in the editor which references an internal path with spaces in it (or other characters that are encoded in URLs, like '(' or ')' ), and you save the node multiple times, you'll notice that the '%' character which was used to encode the other characters also get encoded.
So, how to reproduce it:
- create a node with a link to something like: '/files/some file path.jpg'
- edit the node and re-save without any changes.
- edit the node again and re-save without any changes.
- The link should look now something like: '/files/some%2520files%2520path.jpg'

The issue is in the LinkProcessor::processUrl() method, for the outbound case, when building the path alias with Url::fromUserInput(). So, the first time when the node gets created, the link is properly saved with '/files/some file path.jpeg'. Then, the link gets loaded and the outbound case is called. During this, the Url::fromUserInput() returns this: '/files/some%20file%20path.jpg'. The node then gets saved and now the link is saved with %20 in the database. And from now on, every time the node gets edited, the '%' character gets encoded by Url::fromUserInput() to %25, so the new path will become: '/files/some%2520file%2520path.jpg'

PR #960, opened 2022-02-21
fix: double encoding in the outbound link processing

chindris:
## Package(s) involved

silverback_gutenberg

## Description of changes

Decode the outbound urls during the link processing so that they are not double encoded when the node gets saved.

## Related Issue(s)

[Multiple url encoding in internal links by the gutenberg url processor](https://github.com/AmazeeLabs/silverback-mono/issues/959)

## How has this been tested?

automated tests (kernel).

### DIFF
--- a/packages/composer/amazeelabs/silverback_gutenberg/tests/src/Kernel/LinkProcessorTest.php
+++ b/packages/composer/amazeelabs/silverback_gutenberg/tests/src/Kernel/LinkProcessorTest.php
@@ -161,6 +161,7 @@ class LinkProcessorTest extends MediaKernelTestBase {
         'inbound' => [
           '/unrouted-path' => '/unrouted-path',
           '/de/unrouted-path' => '/de/unrouted-path',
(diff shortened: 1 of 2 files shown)


packages/composer/amazeelabs/silverback_gutenberg/tests/src/Kernel/LinkProcessorTest.php
```diff
@@ -122,6 +122,8 @@ public function testProcessUrl() {
               'en' => '/node/' . $withoutAlias->id(),
               'de' => '/de/node/' . $withoutAlias->id(),
             ],
+          ],
+          [
```
chindris: The automated tests had some wrongly closed array for the outbound cases

Leksat: I pushed a change that simplifies this monster :)

Leksat (approved):

Issue #827, opened 2021-09-30
Create Gutenberg plugin for Entity Usage module


Leksat:
## What can go wrong with the current Drupal-Gatsby connection

- Create a Gutenberg page with a media block.
- Delete the used media entity.

Gatsby will delete the media object from its database. But the reference will still be there in the Gutenberg page. Which can cause Gatsby builds to fail.

## How to solve

Drupal's [Entity Usage module](https://www.drupal.org/project/entity_usage) provides some warning messages for editors ([example](https://git.drupalcode.org/project/entity_usage/-/blob/8.x-2.0-beta5/entity_usage.module#L132)) which could help prevent the described problems.

As the first step, we need to create a plugin for the module which will count
- media entities embedded to Gutenberg pages
- links to Drupal entities added via Gutenberg

PR #970, opened 2022-03-02
Entity usage plugin for media references in Gutenberg

chindris:
## Package(s) involved

silverback_gutenberg

## Description of changes

Adds en Entity Usage plugin for media embedded in Gutenberg.

## Related Issue(s)

[Create Gutenberg plugin for Entity Usage module](https://github.com/AmazeeLabs/silverback-mono/issues/827)

## How has this been tested?

Locally, manual.


pmelab (changes requested):
Thanks! The implementation looks on point. But I think we should have a basic test that makes sure it still works. When for example the block parser changes and its not called `mediaEntityIds` any more.

Here are the other tests implement by @Leksat :
https://github.com/AmazeeLabs/silverback-mono/tree/development/packages/tests/silverback-gutenberg/playwright-tests

chindris:
> Thanks! The implementation looks on point. But I think we should have a basic test that makes sure it still works. When for example the block parser changes and its not called `mediaEntityIds` any more.
> 
> Here are the other tests implement by @Leksat : https://github.com/AmazeeLabs/silverback-mono/tree/development/packages/tests/silverback-gutenberg/playwright-tests

Hi @pmelab !
I added a set of playwright tests and also updated the silverback-drupal app to contain the entity usage module.

pmelab (approved):

pmelab:
Awesome! Thanks @chindris !

Leksat (approved):

Issue #853, opened 2021-10-21
Gatsby Interface translations in Drupal


pmelab:
Store Gatsby interface translations in Drupal to simplify the translation process.

- [x] use the formatjs scanner to extract translations during `prepare` into a file
- [x] import new translations from this file on lagoon deployment
- [x] expose translation strings (maybe groups?) as `silverback_gatsby` streams
- [x] fetch translations during gatsby build and feed them into react-intl

Additionally:
- [x] add eslint-plugin-formatjs
- [x] add eslint rules that stop us from leaving hardcoded strings in react components

PR #996, opened 2022-03-21
Interface translations

chindris:
## Package(s) involved

silverback_gatsby
silverback_translations

## Description of changes
String translation sources can be created via a REST API endpoint, using a JSON string generated by the [formatjs](https://formatjs.io/) library. Translations are as well exposed via a silverback_gatsby stream so they can be consumed by a gatsby app which can inject them into the react-intl translation provider.

## How it works
The silverback-gatsby app has now, as an example, the text in the header of the table for the homepage and article translated. With the `yarn translations:push_sources` command, the source strings are pushed to the drupal app. Then, in the drupal app, you can filter the strings by the `gatsby` context to see all the available entries. You can translated them to the different languages, and the next gatsby build should pull the translations into the gatsby app.

## Related Issue(s)
[Gatsby Interface translations in Drupal](https://github.com/AmazeeLabs/silverback-mono/issues/853)

## How has this been tested?

Locally, manual.


Leksat (commented):
Wow. This is a lot of code 😅 

Can we have at least some tests for it? Or are they planned for later?

chindris:
> Wow. This is a lot of code 😅
> 
> Can we have at least some tests for it? Or are they planned for later?

Hi @Leksat ! yes, we definitely should have tests for that, and most probably it will be the next thing I will do. But we need this feature on a project, that is why I think we need this PR to be merged pretty soon.

Leksat:
Ok, let's make the current tests green, then merge it and... 🤞 😅

chindris:
@Leksat @pmelab we have the tests green now.

There were multiple issues, but basically the main problem was that, because a lot of string sources get created when drupal is installed or for example when the javascript files get parsed for t() calls (and the drupal gutenberg translations js file has thousands of them), and basically each of them had to be tracked by the gatsby feed. So, I did two main improvements in this area:
- when getting the extensions in the ComposableSchema class: https://github.com/AmazeeLabs/silverback-mono/pull/996/commits/45c268b87a74a46c8fe97472a335882165c91a23 they are now statically cached. The problem is that during getting the extensions, a call to getSchemaDefinition() is done, and that one scans the file system for a .graphqls files and return their content. The problem: as stated above, during some requests there might be thousands of calls to this function, which means the file system has to be accessed thousands of times, and that took in some cases on my local computer even more than 70 seconds.
- to avoid a lot of unnecessary calls, we for now only track a string if it has a context: https://github.com/AmazeeLabs/silverback-mono/pull/996/files#diff-5cc15af51aa41c945b1435dfc1bc51eb0d1f98152f1abc360ec7ad02f5a0082dR84 . We need to provide a context anyways when the strings are created using the endpoint, so it makes sense to not track the other ones. However, I'm still not 100% sure we won't need strings with no context exposed in the future, but when we do it we can spend some time trying to improve that part as well.

There are also many exported config files because of translations.

Leksat:
@chindris Great job!

Noticed `drupal_static` which is deprecated: https://www.drupal.org/node/2661204

Will take another look at the whole PR later.

apps/silverback-gatsby/scripts/push_translation_sources.ts
```diff
@@ -0,0 +1,44 @@
+import axios from 'axios';
+import { existsSync,readFileSync } from 'fs';
```
Leksat: Tip: Prettier on Save in PhpStorm
Screen Shot 2022-04-26 at 09 03 24

chindris:
> @chindris Great job!
> 
> Noticed `drupal_static` which is deprecated: https://www.drupal.org/node/2661204
> 
> Will take another look at the whole PR later.

I know about the deprecation from https://www.drupal.org/node/2661204 but the problem is that the class member does not work, I suppose because the object is re-created every time... I'll think of something else.

packages/composer/amazeelabs/silverback_gatsby/graphql/string_translation.directive.graphqls
```diff
@@ -0,0 +1,7 @@
+# Directive for the "StringTranslationFeed" plugin.
+directive @string_translation(
```
Leksat: Why not camelCase?

Other directives are camelCase:  https://github.com/AmazeeLabs/silverback-mono/blob/d0b45ef682bccaecc7696a66256ddd9b037dee4d/packages/composer/amazeelabs/silverback_gatsby/graphql/silverback_gatsby.base.graphqls#L33-L35

apps/silverback-drupal/web/modules/custom/silverback_gatsby_test/graphql/silverback_gatsby_test.graphqls
```diff
@@ -91,3 +91,6 @@ type MenuItem {
 type MainMenu @menu(menu_id: "main", item_type: "MenuItem")
 
 type FirstLevelMainMenu @menu(menu_id: "main", item_type: "MenuItem", max_level: 1)
+
+type GatsbyStringTranslation @string_translation(context_prefix: "gatsby")
+type WebsiteStringTranslation @string_translation(context_prefix: "website")
```
Leksat: Forgotten?

packages/composer/amazeelabs/silverback_gatsby/graphql/string_translation.directive.graphqls
```diff
@@ -0,0 +1,7 @@
...
+  """
+  The string context
```
Leksat: I'd either remove the comment or let it say "context _prefix_".

packages/composer/amazeelabs/silverback_gatsby/graphql/string_translation.directive.graphqls
```diff
@@ -0,0 +1,7 @@
...
+  """
+  context_prefix: String
```
Leksat: camelCase too?

Leksat (approved):
My silly comments as promised 😄

Leksat:
@pmelab FYI we found a trick for the case if the same string needs to have different translations in different places.

You just give it a custom ID
Screen Shot 2022-04-26 at 11 26 30

And then it's visible in the Drupal UI
Screen Shot 2022-04-26 at 11 18 01

pmelab (approved):
This is awesome! 🥳 

As already mentioned, I doubt that we need translation strings without context. In the future we might rather have to be able to drill it down to only strings within certain contexts, if another module creates a lot of strings in its own context that we don't need.

pmelab:
@Leksat : Currently we have an ESLint rule to prevent that trick. And we could run into situations (copy & paste error), were the same id's are used accidentially. Thats why the recommendation is to autocreate them. `formatMessage` has an "description" input for that case. We could create our own function to generate id's, that takes that description and adds the generated id in parenthesis or something. But that would happen in the project setup, not here.

pmelab:
Additional steps to be taken after discussion between @chindris @Leksat @nbaosullivan and @pmelab 

* ignore the generated id when importing into drupal, use the defaultMessage as source string and don't create a first translation
* add the `description` to the context `gatsby: [description]`
* use source strings and description without prefix (`[description]`) to map extracted keys to drupal translation api output to feed it into `IntlProvider`


potential next steps:
* use `--extract-locations` to add location information (typescript file) to drupal translation information
* enhance drupal translation ui to display links to storybook
* enhance drupal translation ui to search for partial contexts
* default-filter for the gatsby context when displaying the translations list in drupal

pmelab (changes requested):
See comment above

pmelab (approved):

PR #1025, opened 2022-04-20
feat: add write-cache command

Leksat:
## Package(s) involved

`silverback-cli`

## Description of changes

A new command to write current Drupal state into `install-cache.zip`.

## Motivation and context

The default content import takes too long.

So now we put the default content into the install cache. And a usual `drupal-install` script looks something like this
```
silverback setup --profile minimal && content:import && silverback update-cache
``` 

And the usual setup will be faster, because we have [Do not reimport existing entities](https://www.drupal.org/project/default_content/issues/2698425) patch in our `proxy-default-content` package.

## How has this been tested?

On a project.

### DIFF
--- a/packages/composer/amazeelabs/silverback-cli/src/AmazeeLabs/Silverback/Commands/WriteCache.php
+++ b/packages/composer/amazeelabs/silverback-cli/src/AmazeeLabs/Silverback/Commands/WriteCache.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace AmazeeLabs\Silverback\Commands;
+
+use Alchemy\Zippy\Zippy;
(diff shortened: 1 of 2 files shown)


pmelab:
I think this was there at some point and you said "it has to many commands, we should remove it" 😄

Leksat:
Checked Git history. We never had it 😛

pmelab:
Other repository! https://github.com/AmazeeLabs/silverback-cli/blob/master/src/AmazeeLabs/Silverback/Commands/SnapshotCreate.php

Not exactly the same, but similar.

Leksat:
`snapshot-create` still exists and is actively used 😉

PR #1037, opened 2022-04-28
feat(gatsby): more flexible and explicit @menu directives

pmelab:
## Package(s) involved

`@amazeelabs/silverback_gatsby`

## Description of changes

* Deprecated the `item_type` declaration and magic creation of `items`, `id`, and `parent` fields in favor of a more explicit way to declare menu types.
* Added directives for resolving items of a menu, urls and labels.

## Motivation and context

Make menu schemas less magic, and more declarative and therefore flexible. With this approach, we can easily change field names or use other resolvers if necessary. Also the schema syntax is not broken upfront and the schema contains the whole type definition.

Instead of:

```graphql
type MainMenu @menu(menu_id: "main", item_type: "MenuItem") {
}

type MenuItem {
   label
   url
}
```

We write:

```graphql
type MainMenu @menu(menu_id: "main") {
  items: [MenuItem!]! @resolveMenuItems
}

type MenuItem {
  id: String! @resolveMenuItemId
  parent: String @resolveMenuItemParentId
  label: String! @resolveMenuItemLabel
  url: String! @resolveMenuItemUrl
}
```

... and there is no need for extra resolvers any more.

### DIFF
--- a/packages/composer/amazeelabs/silverback_gatsby/modules/silverback_gatsby_example/graphql/silverback_gatsby_example.graphqls
+++ b/packages/composer/amazeelabs/silverback_gatsby/modules/silverback_gatsby_example/graphql/silverback_gatsby_example.graphqls
@@ -16,16 +16,22 @@ type Post @entity(type: "node", bundle: "blog") {
 }
 
 type MenuItem {
-  label: String!
-  url: String!
+  id: String! @resolveMenuItemId
+  parent: String @resolveMenuItemParentId
+  label: String! @resolveMenuItemLabel
+  url: String! @resolveMenuItemUrl
 }
 
 """
 All menu items, for the sitemap or being post-loaded.
 """
-type MainMenu @menu(menu_ids: ["access_denied", "main"], item_type: "MenuItem")
+type MainMenu @menu(menu_ids: ["access_denied", "main"]) {
+  items: [MenuItem!]! @resolveMenuItems
+}
 
 """
 Menu items up to level 2, to be rendered directly into the page layout.
 """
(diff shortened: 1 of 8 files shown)


packages/composer/amazeelabs/silverback_gatsby/modules/silverback_gatsby_example/graphql/silverback_gatsby_example.graphqls
```diff
@@ -16,16 +16,22 @@ type Post @entity(type: "node", bundle: "blog") {
...
+  label: String! @resolveMenuItemLabel
+  url: String! @resolveMenuItemUrl
 }
 
 """
 All menu items, for the sitemap or being post-loaded.
 """
-type MainMenu @menu(menu_ids: ["access_denied", "main"], item_type: "MenuItem")
+type MainMenu @menu(menu_ids: ["access_denied", "main"]) {
+  items: [MenuItem!]! @resolveMenuItems
```
Leksat, line 29: Can `!` in `MenuItem!` be a source of issues on Gatsby side?

Leksat (approved):
It's a breaking change, right? => No

packages/composer/amazeelabs/silverback_gatsby/modules/silverback_gatsby_example/graphql/silverback_gatsby_example.graphqls
Leksat: It was mandatory previously and we had no issues with it.

PR #1049, opened 2022-05-06
feat(gatsby): graphql directives for gutenberg documents

pmelab:
## Package(s) involved

## Description of changes

Add directive for efficient Gutenberg GraphQL API's.

### Directive Definitions
https://github.com/AmazeeLabs/silverback-mono/blob/cfda343a6bcdd32c44ed368f63a596d429107991/packages/composer/amazeelabs/silverback_gatsby/graphql/editor_block.directive.graphqls#L1-L31

### Example usage from tests
https://github.com/AmazeeLabs/silverback-mono/blob/cfda343a6bcdd32c44ed368f63a596d429107991/packages/composer/amazeelabs/silverback_gatsby/modules/silverback_gatsby_example/graphql/silverback_gatsby_example.graphqls#L7-L31

## Motivation and context

Write less resolvers for Gutenberg blocks.

## How has this been tested?

With Kerneltests.


packages/composer/amazeelabs/silverback_gatsby/modules/silverback_gatsby_example/graphql/silverback_gatsby_example.graphqls
```diff
@@ -4,9 +4,32 @@ schema {
 
 type Query
 
+union WithPath = Page | Post
```
Leksat, line 7: Forgotten?

packages/composer/amazeelabs/silverback_gatsby/tests/src/Kernel/EditorBlocksTest.php
```diff
@@ -0,0 +1,160 @@
...
+        ],
+      ],
+    ]);
+    $media->addTranslation('de', [
+      'bundle' => 'image',
+      'name' => 'Screaming hairy armadillo',
+      'field_media_image' => [
+        [
+          'target_id' => 1,
+          'alt' => 'Screaming hairy armadillo DE',
```
Leksat, line 53: `Hairy armadillo screaming in German`?

Leksat (approved):
Looks awesome!

How about using the new directives on the existing `GutenbergPage` type?
https://github.com/AmazeeLabs/silverback-mono/blob/c658e5c65e8a3cd2a9b317d93f07ff84607fd772/apps/silverback-drupal/web/modules/custom/silverback_gatsby_test/graphql/silverback_gatsby_test.graphqls#L36-L40

There are Playwright tests for this type. We could double check if the link processing works right.

Leksat:
Hmm, maybe it's too late but... Block directives were added to `silverback_gatsby` module, yet they have no relation to Gatsby. Maybe `silverback_gutenberg` was the right place?

packages/composer/amazeelabs/silverback_gatsby/modules/silverback_gatsby_example/graphql/silverback_gatsby_example.graphqls
pmelab: Added that intentionally, since i broke non-block unions at some point.

packages/composer/amazeelabs/silverback_gatsby/tests/src/Kernel/EditorBlocksTest.php
pmelab: 😆 copied that one from some drupal core unit test

pmelab:
> Hmm, maybe it's too late but... Block directives were added to `silverback_gatsby` module, yet they have no relation to Gatsby. Maybe `silverback_gutenberg` was the right place?

My intention was to make directives pluggable in a next step, so we can easily put them where they belong.

pmelab:
@Leksat changing the test schema was a good call, since it brought more edge cases:

* I added an option to "ignore" blocks during processing. `core/group` is ignored by default. Children of ignored blocks are spread into the current block list.
* `BlockHtml` was spread into 3 different block types. I also thought about allowing multiple block types to be unified, but i felt like this should be the applications decision if a list is the same as a quote is the same as a paragraph.

Leksat:
Super awesome! Let's see what tests will say :)

Leksat (approved):

Issue #1068, opened 2022-05-20
EX: Periodic builds


pmelab:
Instead of building on every content change, trigger builds based on cron or on request.

- run builds on configurable cron
- allow editors with appropriate permissions to trigger a build manually
- use information from `build.json` provided by gatsby to only trigger a build if necessary

PR #1072, opened 2022-05-24
feat: periodic build

colorfield:
## Package(s) involved

silverback_gatsby (Composer package)

## Description of changes

- Makes the build on entity save optional
- Adds a Drush command that triggers the Gatsby build
- Adds a Gatsby Build button
- Moves build specific configuration in its own tab

## Motivation and context

Feature request

## Related Issue(s)

#1068 

## How has this been tested?

- Locally, manual test
- Unit test for `GatsbyUpdateTrackerTest::testDeferredBuild()`


colorfield:
- ~~Gatsby Build button still needs to be exposed via a separate permission~~
- ~~Default value for `build_trigger_on_save` needs to be set to true by default to preserve original behaviour~~

pmelab (commented):
👏

Issue #1060, opened 2022-05-20
EX: Display latest version of content if permitted


pmelab:
If an editor changes a page and clicks the preview button, they should see their changes immediately, even if the build has not finished or even started.

- cross domain cookies have been a problem in the past. maybe routing api requests through netlify to drupal and therefore use the same domain is a better option.

Issue #1062, opened 2022-05-20
EX: Show live/preview version on node page in Drupal


pmelab:
On Drupal node pages, the iframe with the current live or preview version of the page should be displayed.

- Make it possible to enable/disable loading of latest version (e.g. GET-parameter)
- Replicate "latest revision" tab when content moderation is in use.

colorfield:
This will be integrated in `silverback_external_preview` package.

@pmelab, a summary of the discussed UX topics
- On the node 'View' tab, we could have a selector to view the current live or preview version (to append the `preview` GET parameter)
- As the 'Latest revision' tab will show only non default revision, we could skip this selector and always use preview
- When the preview buttons are opened, they are hiding the Save button on Claro, so until we have auto-save it requires to collapse the buttons to hit save. Current resolution could be to delegate this to the Drupal behavior that redirects to 'View' / 'Latest revision' tabs on save.

colorfield:
@pmelab view tab integration is done here https://github.com/AmazeeLabs/silverback-mono/commit/769444d5996071135b6c1d258b673e7ead38b927, i will wait for https://github.com/AmazeeLabs/silverback-mono/pull/1085 to be merged to add the last revision tab support

PR #1085, opened 2022-06-19
feat(gatsby): graphql fields for fetching entity revisions

pmelab:
## Package(s) involved

## Description of changes

Add `load[Type]Revision` root level fields to retrieve specific revisions of an entity

## Motivation and context

Can be used to display specific revisions when loading them in the browser for preview purposes.

## Related Issue(s)

#1060 

## How has this been tested?

Kerneltests

### DIFF
--- a/apps/silverback-drupal/generated/silverback_gatsby.composed.graphqls
+++ b/apps/silverback-drupal/generated/silverback_gatsby.composed.graphqls
@@ -307,6 +307,7 @@ extend type Image {
   translations: [Image!]!
 }
 
+extend type Query { loadImageRevision(id: String!, revision: String!) : Image }
 extend type Query {
   loadTag(id: String!): Tag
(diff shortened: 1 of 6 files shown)


Leksat (approved):
Looks good!

How would client fetch the list of available revisions?

colorfield (approved):

colorfield:
@Leksat this PR will unblock the implementation of the last revision tab in https://github.com/AmazeeLabs/silverback-mono/issues/1062, we could perhaps create a follow-up for the list of available revisions?

pmelab:
@Leksat listing all revisions is not a requirement at the moment. we need this so Drupal can trigger preview of one specific revision (e.g. from Drupals revision list).

Leksat:
Ah! So it will be something like `/page-template?id=123&revision=12345` path?

pmelab:
@Leksat precisely

PR #1087, opened 2022-06-21
feat: add ability to configure intl in storybook

Leksat:
Noticed on a project that `intl` is using `en` locale for _pages_ in Storybook. While other regular components were using `de` locale, as it was set in `IntlDecorator`.


Leksat:
@pmelab I maybe did something wrong here as `IntlDecorator` already wraps everything including pages.

PR #1094, opened 2022-06-29
fix: php fatal error because on invalid url argument

chindris:
## Package(s) involved

silverback_external_preview

## Description of changes

Fix for a fatal error because of an invalid function argument when creating a Url object.
The error: "Auf der Website ist ein unvorhergesehener Fehler aufgetreten. Bitte versuchen Sie es später nochmal."
And from the logs table: "

InvalidArgumentException: The URI '/de/news/test-after-update' is invalid. You must use a valid URI scheme. in Drupal\Core\Url::fromUri() (Zeile 293 in /Users/vasile.chindris/Sites/nuklear9/apps/cms/web/core/lib/Drupal/Core/Url.php).
--"


chindris:
Closing the PR since the problem was that some env variables were not set.

Issue #1061, opened 2022-05-20
DX: Auto-generate gatsby fragments from *.gql files


pmelab:
Instead of maintaining the same fragment twice (e.g. `Page.fragment.ts` and `Page.gql`), a script transforms the `*.gql` file into the gatsby equivalent.

Script could run on:

- prepare
- phpstorm file watcher
- ...?

PR #1097, opened 2022-07-05
feat: auto-generate gatsby fragments from *.gql files

colorfield:
## Package(s) involved

`gatsby-fragments` (new package)

## Description of changes

Bundles fragments generation as a CLI utility

## Related Issue(s)

#1061 

## How has this been tested?

Automated tests.


packages/npm/@amazeelabs/gatsby-fragments/rollup.config.js
```diff
@@ -0,0 +1,18 @@
...
+  input: 'src/index.ts',
+  output: {
+    file: 'index.js',
+    format: 'cjs',
+  },
+  plugins: [
+    nodeResolve({ preferBuiltins: true }),
+    typescript(),
+    commonjs(),
+    json(),
```
pmelab: Why do we need JSON resolution? Do you import a `*.json` file anywhere? I could not find anything.

packages/npm/@amazeelabs/gatsby-fragments/src/generate.test.ts
```diff
@@ -0,0 +1,38 @@
+import { readFileSync } from 'fs';
+import { sync } from 'glob';
+
+import { generate } from './generate';
+
+describe('generate', () => {
+  const filesAmountEquality = () => sync('./test/*.gql').length === sync('./test/*.fragment.ts').length;
```
pmelab: Did you consider using [mock-fs](https://www.npmjs.com/package/mock-fs)? Using the filesystem in jest can be dangerous because we could run tests in parallel. Also I can make tests more concise.

pmelab (changes requested):
That looks great! Just two small remarks 😄

packages/npm/@amazeelabs/gatsby-fragments/rollup.config.js
colorfield: Thank you for spotting, this is a leftover, removing it.

colorfield: Removed in https://github.com/AmazeeLabs/silverback-mono/commit/ed2e2027e730737a113d0e31d3f2f620c9b1762b

pmelab (approved):

packages/npm/@amazeelabs/gatsby-fragments/src/generate.test.ts
colorfield: Running now tests in isolation with mock-fs https://github.com/AmazeeLabs/silverback-mono/commit/c44f3e9eb6210b8eb210789e008fc77a7f5c5204. 
Also added async/await so the test would stay valid if we decide to use `writeFile` and `readFile` in `generate` instead of `readFileSync` and `writeFileSync`.

PR #1100, opened 2022-07-07
fix: add timeout to graphql executor

Leksat:
## Package(s) involved

`@amazeelabs/gatsby-source-silverback`

## Description of changes

Added a one minute timeout to GraphQL requests.

If a request cannot be finished in the given time, `gatsby build` will fail. At least this is the behavior I saw locally when simulated endless requests.

## Motivation and context

If a request gets stuck for some reason, the build never finishes.

## How has this been tested?

Tested locally. Works.


Leksat:
@pmelab @colorfield Do you think one minute is okay?

pmelab (approved):

Issue #1103, opened 2022-07-11
Make the external preview computed field private


colorfield:
The external preview computed field can be rendered to anonymous users by e.g. Search API and thus display the output provided by the formatter.

colorfield:
Closed with #1104

PR #1104, opened 2022-07-11
fix: make external preview computed field private by default

colorfield:
## Package(s) involved

`silverback_external_preview`

## Description of changes

Makes the external preview computed field private by default.
By extension, as we are introducing a new permission, also use it for the toolbar.
Initially, i was thinking about checking if the user is authenticated, but there could be use cases where we want to use a permission for that (e.g. webforms for authenticated users).

## Motivation and context

Can be displayed on the frontend when using entity Drupal rendering.

## Related Issue(s)

#1103 

## How has this been tested?

Manually.


pmelab (approved):
👌

PR #1109, opened 2022-07-14
fix: include unpublished entities in the diff

colorfield:
## Package(s) involved

`silverback_gatsby`

## Description of changes

Add, to the entity access condition, a condition that checks if the entity is publishable and if it is unpublished.

## Motivation and context

Unpublishing an entity can lead to not have it in the diff for Gatsby sourcing, so it will not increment the drupalBuildId and not be part of the changes. This can be checked with.

```
{
  drupalBuildId
  drupalFeedInfo {
    typeName
    changes(lastBuild: [last_id], currentBuild: [current_id])
  }
}
```
## How has this been tested?

Manually.

### DIFF
--- a/packages/composer/amazeelabs/silverback_gatsby/src/Plugin/Gatsby/Feed/EntityFeed.php
+++ b/packages/composer/amazeelabs/silverback_gatsby/src/Plugin/Gatsby/Feed/EntityFeed.php
@@ -4,6 +4,7 @@ namespace Drupal\silverback_gatsby\Plugin\Gatsby\Feed;
 
 use Drupal\content_translation\ContentTranslationManagerInterface;
 use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Entity\EntityPublishedInterface;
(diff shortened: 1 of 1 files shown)


colorfield:
This case is covered by https://github.com/AmazeeLabs/silverback-mono/blob/development/packages/composer/amazeelabs/silverback_gatsby/silverback_gatsby.module#L22-L25 and does not happen on a clean Drupal 9.4.2. This needs some more debug to figure out if this a core or project specific issue.

pmelab:
@colorfield i think @Leksat also encountered this, and it was related to the node access table when more complex permission systems are in place. I would classify this as "project specific".

Leksat:
What I met was fixed in https://github.com/AmazeeLabs/silverback-mono/pull/1161. @colorfield maybe check if `amazeelabs/silverback_gatsby` version is less than 1.25.1 on your project?

colorfield:
Thanks @Leksat it was indeed < 1.25.1, going to remove the patch and upgrade to test if it is the same issue.

Issue #1067, opened 2022-05-20
EX: CSR pages for all content types


pmelab:
The `cdn_redirect` module can detect requests to pages that should be available, but are not built yet. These are rewritten to a `___csr/[content-type]` template that will load data based on the current path. This way we can have a preview of a new page even before it was build.

- [x] load by path functionality #1149 
- [ ] create a template for each content type that takes the current path and attempts to "live" resolve it
- [x] add logic to `cdn_redirect` that detects that a page that netlify tries to redirect "should" exist, and proxy the types preview template instead (like the current 404 proxy) #1152 
- [x] reduce cache max age for redirect and 404 responses to 5 minutes #1148

PR #1149, opened 2022-08-17
feat(gatsby): load entities by path

pmelab:
Allow to load entities by path to facilitate direct routing.

## Package(s) involved

## Description of changes

The `id` argument now also allows paths, in addition to ids and uuids. If the `id` starts with a `/`, the resolver attempts to find the corresponding entity and resolves that instead. Language and revision loading remains the same, the path only replaces the id.

## Motivation and context

Load entities in the front-end by their assigned path, for preview or dynamic displays.

## Related Issue(s)

#1067 

## How has this been tested?

* Kerneltest
* Manual test in a Drupal instance

### DIFF
--- a/packages/composer/amazeelabs/silverback_gatsby/src/Plugin/GraphQL/DataProducer/FetchEntity.php
+++ b/packages/composer/amazeelabs/silverback_gatsby/src/Plugin/GraphQL/DataProducer/FetchEntity.php
@@ -6,12 +6,15 @@ use Drupal\Core\Entity\EntityRepositoryInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Entity\RevisionableInterface;
 use Drupal\Core\Entity\TranslatableInterface;
+use Drupal\Core\Path\PathValidator;
+use Drupal\Core\Path\PathValidatorInterface;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\graphql\GraphQL\Buffers\EntityBuffer;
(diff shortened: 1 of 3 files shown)


packages/composer/amazeelabs/silverback_gatsby/src/Plugin/GraphQL/DataProducer/FetchEntity.php
```diff
@@ -6,12 +6,15 @@
...
 use Drupal\Core\Entity\RevisionableInterface;
 use Drupal\Core\Entity\TranslatableInterface;
+use Drupal\Core\Path\PathValidator;
+use Drupal\Core\Path\PathValidatorInterface;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\graphql\GraphQL\Buffers\EntityBuffer;
 use Drupal\graphql\GraphQL\Buffers\EntityRevisionBuffer;
 use Drupal\graphql\GraphQL\Execution\FieldContext;
 use Drupal\graphql\Plugin\GraphQL\DataProducer\DataProducerPluginBase;
```
colorfield: nit: some imports might not be needed
- `use Drupal\Core\Entity\RevisionableInterface;`
- `use Drupal\Core\Path\PathValidator;` (using PathValidatorInterface)
- `use Drupal\redirect\RedirectRepository;`

colorfield:
There is a failing test due to a timeout, but cannot reproduce locally, so we might just re-run it (see https://www.githubstatus.com/incidents/2nd6fyb4drjx)

Capture d’écran 2022-08-17 à 19 04 15

PR #1150, opened 2022-08-17
feat(gatsby): user login status

pmelab:
Provide a graphql field that exposes the current users id and name.

## Package(s) involved

`amazeelabsl/silverback_gatsby`

## Description of changes

Add default graphql fields for retrieving the current user login status.

## Motivation and context

Determine in the frontend if the current user is logged in.

## Related Issue(s)

#1067

### DIFF
--- a/packages/composer/amazeelabs/silverback_gatsby/tests/src/Kernel/CurrentUserTest.php
+++ b/packages/composer/amazeelabs/silverback_gatsby/tests/src/Kernel/CurrentUserTest.php
@@ -8,8 +8,9 @@ class CurrentUserTest extends EntityFeedTestBase {
   function testAnonymousUser() {
     $query = $this->getQueryFromFile('current-user.gql');
     $metadata = $this->defaultCacheMetaData();
-    $metadata->addCacheTags(['node_list']);
+    $metadata->setCacheMaxAge(0);
     $anonymous = User::load(0);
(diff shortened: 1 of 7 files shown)


packages/composer/amazeelabs/silverback_gatsby/tests/src/Kernel/CurrentUserTest.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+namespace Drupal\Tests\silverback_gatsby\Kernel;
+
+use Drupal\user\Entity\User;
+
+class CurrentUserTest extends EntityFeedTestBase {
+  function testAnonymousUser() {
+    $query = $this->getQueryFromFile('current-user.gql');
+    $metadata = $this->defaultCacheMetaData();
```
colorfield, line 10: PR looks good, we can probably change this to make the test pass

```
function testAnonymousUser() {
    $query = $this->getQueryFromFile('current-user.gql');
    $anonymous = User::load(0);
    $metadata = $this->defaultCacheMetaData();
    $metadata->setCacheMaxAge(0);
    $metadata->addCacheTags($anonymous->getCacheTags());
    $this->setCurrentUser($anonymous);
    $this->assertResults($query, [], [
      'currentUser' => [
        'id' => '0',
        'name' => null,
      ],
    ], $metadata);
  }
```

PR #1156, opened 2022-08-18
Use GET requests for persisted queries

Leksat:
## Package(s) involved

`@amazeelabs/silverback-graphql-persisted`

## Description of changes

Use `GET` requests for persisted queries. Persisted mutations should still use `POST` requests.

## How has this been tested?

Playwright tests.


Leksat:
@pmelab @colorfield (other lucky person) If you take it over:

tl;dr: merge it!

The new persisted query package is used only in one project. So it's safe to merge this PR. It won't break anything else 😛 

The new Playwright test uses two new GraphQL pages:
- http://localhost:8000/react-query-test for `react-query`
- http://localhost:8000/graphql-request-test for `graphql-request`

Both pages run a query and a mutation using the autogenerated hooks or sdk.

The test is very basic. It just checks if data is loaded. We could improve it (check network requests to ensure `GET` for queries and `POST` for mutations; add cases for query and mutation without variables; redo the test completely because I think it sucks) but we are very busy, right? 😅 

Anyways, I think it should just work 🤞

PR #1175, opened 2022-09-06
feat(gatsby)!: add page path to CSR template

Leksat:
## Package(s) involved

`amazeelabs/silverback_cdn_redirect`

## Motivation and context

```ts
  args.actions.createPage({
    path: '/___csr',
    matchPath: '/*', // <= 😱 
    component: path.resolve(path.resolve(__dirname, 'src/templates/csr.tsx')),
  });
```

The `matchPath` option makes Gatsby rewrite `/___csr` path to the actual page path. Yet it also kills cdn-redirect completely. Every page unknown to Gatsby becomes a CSR page, 404 pages and other redirects are gone.

## Description of changes

Now we provided the actual page path to CSR templates and can use it from there to override the URL without using `matchPath`.

## How has this been tested?

Extended existing tests.


Leksat:
@pmelab FYI

PR #1208, opened 2022-10-09
feat(gatsby): codegen plugin for persisted query maps and typed ids

pmelab:
## Package(s) involved

`@amazeelabs/codegen-operation-ids`

## Description of changes

Add a new codegen plugin that solves persisted queries and query typing in a framework independent way. Refer to the `README.md` for a detailed explanation.

## Motivation and context

The current approach has the following problems:

* other codegen plugins couple typing directly with api calls (e.g. react-query) which results in inflexible typing and lots of duplicated code
* query maps and full queries are loaded into the client. in combination with preview queries, these can get pretty large.

## How has this been tested?

* unit tests
* integration tests


apps/silverback-gatsby/src/pages/react-query-test.tsx
```diff
@@ -1,15 +1,78 @@
...
   const [randomInt, setRandomInt] = useState<number>();
-  const { data: pages } = useGetPagesQuery({ offset: 0, limit: 2 });
-  const { mutate } = useGetRandomIntMutation({
+  const { data } = usePersistedQuery(GetPagesQuery, {
+    offset: 0,
+    limit: 2,
+  });
+  const { mutate } = usePersistedMutation(GetRandomIntMutation, {
     onSuccess: (data) => {
+      console.log(data);
```
Leksat: A BUUUUUUG 😂

packages/npm/@amazeelabs/codegen-operation-ids/src/index.test.ts
```diff
@@ -0,0 +1,409 @@
...
+//   id: T,
+//   variables: OperationVariables<T>,
+// ): Promise<OperationResult<T>> {
+//   return {};
+// }
+//
+// async function foo() {
+//   const res = await fetcher(PageQuery, { path: '/' });
+//   res.loadPage.title;
+// }
```
Leksat: Forgotten?

Leksat (commented):
1. Should `graphqlFetch`, `usePersistedQuery` and `usePersistedMutation` be moved to a helper package? Or do we want to ~define~ copy-paste them in every project?
2. Could you please mark `@amazeelabs/silverback-graphql-persisted` as deprecated?
3. Can the plugin be used with `amazeelabs/silverback_graphql_persisted`? If yes, maybe let's mention this in the readme? And maybe also update the readme of `amazeelabs/silverback_graphql_persisted` package as it still point to the old solution?

The plugin's code looks looks cool and... magic. I can't say I reviewed it well 😅

pmelab:
> 1. Should `graphqlFetch`, `usePersistedQuery` and `usePersistedMutation` be moved to a helper package? Or do we want to ~define~ copy-paste them in every project?

We could maybe create our own helpers. But i'm not sure if that will even be necessary. Lets see if its needed.

> 
>     2. Could you please mark `@amazeelabs/silverback-graphql-persisted` as deprecated?
> 
>     3. Can the plugin be used with `amazeelabs/silverback_graphql_persisted`? If yes, maybe let's mention this in the readme? And maybe also update the readme of `amazeelabs/silverback_graphql_persisted` package as it still point to the old solution?

I don't see a way to use them together. The new package replaces the old one entirely.

Leksat:
@pmelab `@amazeelabs/silverback-graphql-persisted` !== `amazeelabs/silverback_graphql_persisted` 😉 The latter is a PHP package.

pmelab:
Oh! Sorry, i confused that. Yes, the php package should work immediately.

PR #1212, opened 2022-10-13
Zustand organisms

pmelab:
## Package(s) involved

`@amazeelabs/react-framework-bridge`

## Description of changes

Allow organisms to receive [Zustand](https://github.com/pmndrs/zustand) stores as input.

## Motivation and context

Implement functionality as stores that are also easy to mock in storybook but retain fine grained control what re-renders when.


packages/npm/@amazeelabs/react-framework-bridge/src/utils/__tests__/atomic.test.tsx
```diff
@@ -473,6 +474,121 @@ describe('Route rendering', () => {
...
+          </div>
+          <div>
+            <p>
+              Foo
+            </p>
+          </div>
+        </div>
+      </div>
+    `);
+    vi.runAllTimers();
```
Leksat, line 572: Why it was required to use `setTimeout`? If we called `act(...)` right here, wouldn't it lead to the same result?

PR #1288, opened 2023-02-01
Update preview links

nbaosullivan:
update preview links to use new __preview path on the frontend

BREAKING CHANGE: preview and live hosts are Drupal config settings, rather than ENV vars.

### DIFF
--- a/packages/composer/amazeelabs/silverback_external_preview/README.md
+++ b/packages/composer/amazeelabs/silverback_external_preview/README.md
@@ -1,10 +1,13 @@
 # Silverback external preview
-Provides a UI to preview pages (published and draft) on an external website - usually a Gatsby or NextJS frontend.
 
-## Environment variables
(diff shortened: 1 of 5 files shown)


apps/silverback-drupal/web/modules/custom/gutenberg_custom/gutenberg_custom.info.yml
```diff
@@ -5,4 +5,4 @@ package: Custom
 core: 8.x
 dependencies:
   - gutenberg:gutenberg (>8.x-2.0-beta2)
-core_version_requirement: ^8 || ^9
+core_version_requirement: ^8 || ^9 || ^10
```
pmelab: Does this belong in this PR? It will cause major version jumps in all the other packages.

pmelab (changes requested):
See inline comments

Issue #1074, opened 2022-05-27
Cloudinary data producer


pmelab:
## Goal
Provide an `amazeelabs/silverback_cloudinary` Drupal module, that contains graphql data producer plugin to add responsive image resolution anywhere in the GraphQL schema.

Recommended reads before implementing or using this extension:

- [Responsive Images](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images)
- [Image transformations](https://cloudinary.com/documentation/transformation_reference)
- [Signed on-the-fly transformations](https://cloudinary.com/blog/on_the_fly_image_manipulations_secured_with_signed_urls)
- [Automatic image formats](https://cloudinary.com/documentation/image_optimization#automatic_format_selection_f_auto)

The goal is to allow frontend developers to request responsive image sizes in a GraphQL query.

```graphql
fragment Hero on Page {
  heroImage(config: {
    # Display a 2000/800 header image by default.
    width: 1600,
    height: 800,
    sizes: [
      # For screens smaller than 800px, scale down to 780px width.
      [800, 780]
    ],
    variants: [
      {
        # Use this variant for small portrait displays like phones.
        media: "(max-width: 800px) and (orientation: portrait)",
        # Request a portrait cut instead of landscape.
        width: 800,
        height: 1600,
        # On mobile, text overlays the image, so we tint it a bit.
        # https://cloudinary.com/documentation/transformation_reference
        transform: "co_rgb:000000,e_colorize:60"
      }
    ]
  })
}
```

The result of this query should be a JSON object with all necessary information to render the following responsive image:

```html
<picture>
  <source
   src="https://res.cloudinary.com/demo/image/fetch/c_fill,w_800,h_1600,co_rgb:000000,e_colorize:60/f_auto/https://www.mywebsite.com/sites/default/files/hero.jpeg"
   media="(max-width: 800px) and (orientation: portrait)"
   width="800"
   height="1600"
  />
  <img
    src="https://res.cloudinary.com/demo/image/fetch/c_fill,w_1600,h_800/f_auto/https://www.mywebsite.com/sites/default/files/hero.jpeg"
    srcset="https://res.cloudinary.com/demo/image/fetch/c_fill,w_780,h_390/f_auto/https://www.mywebsite.com/sites/default/files/hero.jpeg 780w"
    width="1600"
    height="800"
    sizes="(max-width: 800px) 780px, 1600px"
    alt="Hero Image"
[...]

PR #1292, opened 2023-02-09
Cloudinary integration

chindris:
## Package(s) involved

npm: @amazeelabs/gatsby-silverback-cloudinary
composer: amazeelabs/silverback_cloudinary

## Description of changes

Integration of the cloudinary service into silverback.
Detailed information about how these packages work can be found in the readme file of them, and the related issue bellow.

## Related Issue(s)

https://github.com/AmazeeLabs/silverback-mono/issues/1074

## How has this been tested?

Unit tests


apps/silverback-drupal/web/modules/custom/silverback_gatsby_test/src/Plugin/GraphQL/Schema/SilverbackGatsbyTestSchema.php
```diff
@@ -57,6 +57,22 @@ public function getResolverRegistry(): ResolverRegistryInterface {
...
+          ->map('image', $builder->fromParent())
+          ->map('config', $builder->fromArgument('config'))
+      )
+    );
+    $addResolver('GutenbergPage.anotherResponsiveImage',
+      $builder->produce('responsive_image')
+        ->map('image', $builder->fromValue('https://dummyimage.com/1000x500/807e08/182196.jpg'))
+        ->map('config', $builder->fromArgument('config'))
+    );
+
```
pmelab: This could be wrapped in a `directive`, since neither the arguments nor the output are going to change.
https://github.com/AmazeeLabs/silverback-mono/blob/development/packages/composer/amazeelabs/graphql_directives/README.md#extending

packages/composer/amazeelabs/silverback_cloudinary/graphql/silverback_cloudinary.base.graphqls
```diff
@@ -0,0 +1,47 @@
...
+}
+
+type ResponsiveImage {
+  src: String
+  srcset: String
+  sizes: String
+  width: Int
+  height: Int
+  sources: [ResponsiveImageSource]
+}
```
pmelab: I don't think this has to be a dedicated GraphQL type. We could just encode the whole structure into a JSON string and have one frontend component that understands it. These properties are taken from the HTML spec directly and are not going to change.

That way projects don't have to deal with this all the time.

packages/composer/amazeelabs/silverback_cloudinary/graphql/silverback_cloudinary.base.graphqls
```diff
@@ -0,0 +1,47 @@
...
+
+input ResponsiveImageConfig {
+  # See above.
+  width: Int!
+  height: Int
+  sizes: [[Int!]!]
+  transform: String
+  # An array of responsive image variants.
+  variants: [ResponsiveImageVariant!]
+}
```
pmelab: If we drop the `variants` property, we could get away without an input type, since all others are primitives that could be passed as field arguments directly. That in turn would - in combination with the comment below - make the schema extension obsolete.

Reasoning behind it - variants have two purposes:

1. Format negotiation - selecting if webp & friends can be used based on the device.
2. Art direction - displaying entirely different images, based on the viewport.

Reason 1. is already solved by cloudinary's automatic format negotiation, and 2. can also be solved on a CSS / component level, where all other responsive behaviour is handled. One could argue that thats the better place to do that anyway.

packages/npm/@amazeelabs/gatsby-silverback-cloudinary/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "devDependencies": {
+    "@amazeelabs/eslint-config": "^1.4.36",
+    "@amazeelabs/jest-preset": "^1.3.40",
+    "@amazeelabs/prettier-config": "^1.1.1",
+    "@types/crypto-js": "^4.1.1",
+    "@types/jest": "29.2.1",
+    "@types/node": "16.18.11",
+    "eslint": "8.26.0",
+    "gatsby": "~3.3.0",
+    "jest": "29.2.2",
```
pmelab: We use vitest nowadays (although `@amazeelabs/scaffold` still has to be updated). Replacing it would simply mean removing jest and its config, installing vitest and adding `import { describe, it, expect } from 'vitest';` to the test file.

pmelab (changes requested):
Thanks for all that work! Thats already in great shape, but I think there were some misunderstandings and we can simplify some parts.

packages/npm/@amazeelabs/gatsby-silverback-cloudinary/package.json
chindris: we should use now vitest

packages/composer/amazeelabs/silverback_cloudinary/graphql/silverback_cloudinary.base.graphqls
chindris: variants have now been dropped, and we don't have anymore new graphql types.

packages/composer/amazeelabs/silverback_cloudinary/graphql/silverback_cloudinary.base.graphqls
chindris: this has been changed so that the returned data is just a json encoded string.

apps/silverback-drupal/web/modules/custom/silverback_gatsby_test/src/Plugin/GraphQL/Schema/SilverbackGatsbyTestSchema.php
chindris: I created the directive, and it can be used, as long as we can pipe an image string into it.

pmelab (approved):

Issue #848, opened 2021-10-21
External preview breaks some forms


pmelab:
Entity forms without paths break due to the external preview module. It should check first if the current entity should have a preview.

chindris:
This was easy to reproduce when accessing a path alias entity (its edit form route)

PR #1297, opened 2023-03-01
Catch the UndefinedLinkTemplateException

chindris:
## Package(s) involved

amazeelabs/silverback_external_preview

## Description of changes

When generating the preview link, if the entity does not have a canonica url (like, for example, Path alias entities), the UndefinedLinkTemplateException is thrown, and a 500 error gets displayed in the browser. This PR catches this exception (as well as the EntityMalformedException one) and will just return NULL in these cases.

## Related issues
https://github.com/AmazeeLabs/silverback-mono/issues/848

## How has this been tested?

Manually.


pmelab (approved):
Great! That one has been there for long 😄

PR #1298, opened 2023-03-14
chore: set optimize autoload to true in composer recipe

dan2k3k4:


Leksat:
I thought recipes are dead. No?

dan2k3k4:
@Leksat mmm unsure, noticed at least one client project was missing the key, and it was based on the Drupal Recipe

PR #1299, opened 2023-03-15
chore: standardise .gitignore to ignore IDE and macOS files

dan2k3k4:


.gitignore
```diff
@@ -1,5 +1,19 @@
+# Node
 node_modules
 .npmrc
 
-# Local Netlify folder
-.netlify
```
Leksat, line 6: Are you sure about this removal?

dan2k3k4: Hmm that was supposed to stay 🙈

.gitignore
```diff
@@ -1,5 +1,23 @@
+# Node
 node_modules
 .npmrc
 
+# JetBrains
+**/.idea/workspace.xml
+**/.idea/tasks.xml
+**/.idea/shelf
```
Leksat: PhpStorm config is already under git: https://github.com/AmazeeLabs/silverback-mono/blob/development/.idea/.gitignore

So I guess we don't need this.

Leksat (changes requested):

Leksat:
@dan2k3k4 Once you have everything done, could you please squash all commits?

.gitignore
```diff
@@ -1,8 +1,13 @@
 _local
-node_modules
 .phpunit.result.cache
 
 # Local Netlify folder
 .netlify
 
+# Node / Deno
+node_modules
+.npmrc
```
Leksat: Why to ignore this one?

dan2k3k4: Hmm, so the `Gitignore` changes came from using `npx gitignore`

I see that we already have a `.npmrc` committed, and we're not really using `npm` anywhere. I don't think `pnpm` reads the `.npmrc` file.

So I've removed it from the branch

Leksat: https://pnpm.io/8.x/npmrc 😉

Leksat (approved):

Issue #1320, opened 2023-03-28
Add context to Silverback Gutenberg validator


colorfield:
## Package

`silverback_gutenberg`

## Description

Scope: backend validator.

Only the field name is displayed when a validation error occurs. Example `Link URL field is required.`. Which can make it challenging to identify the source of the error when there are many blocks.

We could add more context and display the block name that is triggering the validation error. Example `Link List: Link URL field is required.`.

## Possible follow up

Extend with client side validator.

Issue #1293, opened 2023-02-14
Write tests for Gutenberg validators


colorfield:
Related to #1279 

List of tests to produce

## Field level 

* Single rule - example: required
* Multiple rule - example: required and email

## Block level

* Custom validation rule 
* Cardinality validation 
    * Test variations
        * Limit by any block type
        * Limit by a specific  block type 
    * Tests that applies to both variations
        * Limit by min / max
        * Limit by min = max (exactly)
        * Limit by min / unlimited
        * Limit by 0 / max
        * Limit by 0 / unlimited  
    * Tests that applies to the specific block type only
        * Single block child
        * Combine blocks children

colorfield:
Closed by  #1359

PR #1323, opened 2023-03-29
feat(drupal): prefix Gutenberg blocks validator violation message with breadcrumbs

colorfield:
## Package(s) involved

`silverback_gutenberg`

## Description of changes

EX - Adds context to the validation violation to identify which block is triggering the error message.

As we have a similar representation of the DOM with the blocks structure, we could extend this in a follow-up to highlight the block that is triggering the error and scroll to it (e.g. bubble this and pass it as a Drupal setting that can be used by JS).

The breadcrumbs display is based on the blocks machine name as we don’t have the block label in the Gutenberg backend validation context, but it should match the label in most cases (so `link-list` becomes `Link list`).

If we want more flexibility, we need to think about a solution to expose the block label to the backend. Examples:
- set it in each validation plugin but this would require duplication and extra maintenance
- some DOM parsing could also work.

But we might also think about full client side validation as well so this is an MVP to add context.

## Related Issue(s)

#1320 

## How has this been tested?

Manually, on a project, validator tests to be done in #1293 

Example
Capture d’écran 2023-03-28 à 22 18 01

### DIFF
--- a/packages/composer/amazeelabs/silverback_gutenberg/src/Plugin/Validation/Constraint/GutenbergValidator.php
+++ b/packages/composer/amazeelabs/silverback_gutenberg/src/Plugin/Validation/Constraint/GutenbergValidator.php
@@ -109,26 +109,27 @@ class GutenbergValidator extends ConstraintValidator implements ContainerInjecti
    * @param $block
    * @param array $plugins
    */
-  public function validateBlocks(array $blocks, array $plugins) {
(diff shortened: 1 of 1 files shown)


Leksat (commented):
I'm not familiar with the latest Gutenberg work we did, but the screenshot looks promising 😅

pmelab (approved):

PR #1326, opened 2023-03-30
fix: support directives on enums

Leksat:
## Package(s) involved

`amazeelabs/graphql_directives`

## Motivation and context

```
MissingDefaultException: Missing @default directive for type MyEnum. Either add a @default directive to MyEnum or turn all occasions of 'MyEnum!' into 'MyEnum'
```

## How has this been tested?

Extended unit tests.


pmelab (approved):
🥳

PR #1339, opened 2023-04-19
feat(publisher): Authenticate Publisher with Drupal

colorfield:
## Package(s) involved

- Publisher
- Drupal app

## Description of changes

OAuth2 middleware that is checking if a Drupal user has the `Access Publisher` permission to grant access to Express authenticated routes.

Grant types

- `AuthorizationCode`
- `ResourceOwnerPassword`

Keeps BC with existing `Basic Auth`: provides a configurable way, so nothing will change on existing setups unless `OAuth2` is configured.

Opens the possibility to generalise OAuth2 for other use cases: Instant preview, Gatsby sourcing, end user or other APIs authentication.

Documentation for Publisher and Drupal configuration.

### TBD

Drupal `Publisher` scope/role might be a bit too generic and potentially conflict with editorial related roles. Perhaps choose the `Publisher Builder` name.

### Remaining tasks

- Remove default content export for OAuth2 tokens: create a PR against the silverback-template repo that adds this entity type https://github.com/AmazeeLabs/silverback-template/blob/dev/packages/drupal/test_content/export.php
- Default local hash salt needs to be changed to comply with OAuth2 (min 32 chars), so existing default content passwords have to be updated
- Test on a project (deploy on a dev environment)

## Motivation and context

https://amazeelabs.atlassian.net/browse/SLB-176

## Related Issue(s)

https://github.com/AmazeeLabs/silverback-mono/issues/1137

## How has this been tested?

Manually.


pmelab (approved):
Great! Now to apply this in the template!

colorfield:
Going to create a PR in the `silverback-template` repository and reference it here after.

Issue #1349, opened 2023-05-05
Campaign urls


pmelab:
- [ ] A Drupal module that maintains simple redirects from path `/a` to `/b`, without any checks on language or permissions.
- [ ] These redirects should be exported during the Gatsby build process as Netlify redirects that run on edge instead of calling Drupal.
- [ ] It is possible to have the same source path for mulitple domains in multi-domain scenarios

PR #1369, opened 2023-06-06
Campaign urls (Raw redirects)

chindris:
## Package(s) involved

amazeelabs/silverback_campaign_urls
@amazeelabs/gatsby-source-silverback

## Description of changes

This PR introduces a new entity type, called **CampaignUrl**. The backend administrators will be able to create these campaign urls, which are basically just simple redirects, meaning that they are free to input any kind of source and destination paths. There's no domain or language check on them.

The module implements a schema extension plugin which exposes these **CampaignUrl** entities as a Gatsby feed. They are then fetched by the **gatsby-source-silverback** plugin and a redirect is created for each of them.

The entry point for administering the campaign urls (in the backend) is /admin/config/search/campaign_url . From there, people with the **Administer individual campaign urls** permission are able to add / edit /delete them.

## Related Issue(s)

https://github.com/AmazeeLabs/silverback-mono/issues/1349

## How has this been tested?

Locally, manually.


apps/silverback-gatsby/gatsby-node.ts
```diff
@@ -0,0 +1,44 @@
+import { GatsbyNode } from 'gatsby';
```
chindris: @pmelab  should this be extracted maybe in a separate package? It would be easier to use it afterwards in the other apps.

pmelab: Yes. As already discussed we could add a schema extension to the Drupal module that provides the GraphQL type, and Gatsby plugin that executes the query and creates the redirects.

The latter could also be added to `gatsby-source-silverback` 🤔 , so we don't need another plugin. It just has gracefully handle the case when the schema extension is not active. But that information should be in the `FeedInfo`.

pmelab (changes requested):
The module should actually be called `silverback_campaign_urls`. And as discussed in another comment, we can make the remaining Drupal and Gatsby parts reusable as well.

apps/silverback-gatsby/gatsby-node.ts
chindris: Hi @pmelab !

This is now in the `gatsby-source-silverback` package. The `CampaignUrl` entities will generate redirects if the schema extension is enabled in Drupal.

chindris:
> The module should actually be called `silverback_campaign_urls`. And as discussed in another comment, we can make the remaining Drupal and Gatsby parts reusable as well.

The module name has been changed to `silverback_campaign_urls`.

pmelab (approved):
Great! 😊 
I just realized that we should also be able to manage splat rewrites (`/blog/* https://old-blog.com/:splat`) with this 🤯

chindris:
> Great! 😊 I just realized that we should also be able to manage splat rewrites (`/blog/* https://old-blog.com/:splat`) with this 🤯

yes, we can input any source and destination string for the urls, like we'd directly edit the redirects file

PR #1386, opened 2023-06-30
fix: remove dropbutton inline styles

colorfield:
## Package(s) involved

`silverback_publisher_monitor`

## Description of changes

Remove any inline style that can be added by another process on the dropbutton used for the Publisher monitor.
I could not identify yet which process is appending these styles, so the hotfix is to make sure that we don't want any there.

## Motivation and context

Before the fix
Capture d’écran 2023-06-30 à 11 29 03

We have inline styles added at several levels.
Capture d’écran 2023-06-30 à 11 31 15

After

Capture d’écran 2023-06-30 à 12 20 52

## How has this been tested?

Manually, on the `silverback-template` repository.


colorfield:
Stale PR, fixed in a better way on the development branch, closing.

PR #1390, opened 2023-07-05
fix: gutenberg media handling

pmelab:
## Package(s) involved

* `amazeelabs/silverback_gutenberg`

## Description of changes

Change the Gutenberg media resolver to use entity buffer.

## Motivation and context

The initial bug was that `@resolveProperty` fails when chained directly after `@resolveEditorBlockMedia` when the media entity does not exist.

Turns out that `$builder->compose` only aborts on `NULL` values when they are emitted by a `Deferred` result. 

Therefore I changed the EditorBlockMedia resolver to use EntityBuffer (for better performance) and also return a deferred result, which causes the query to return null correctly. Test cases for that have been added.

Another finding was that the `media_list` cache tag was missing up until now, which might have caused caching issues.

- [x] Kernel tests

### DIFF
--- a/packages/composer/amazeelabs/silverback_gutenberg/src/Plugin/GraphQL/DataProducer/EditorBlockMedia.php
+++ b/packages/composer/amazeelabs/silverback_gutenberg/src/Plugin/GraphQL/DataProducer/EditorBlockMedia.php
@@ -3,10 +3,17 @@
 namespace Drupal\silverback_gutenberg\Plugin\GraphQL\DataProducer;
 
 use Drupal\Core\Cache\RefinableCacheableDependencyInterface;
+use Drupal\Core\Entity\EntityRepositoryInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Entity\TranslatableInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\graphql\GraphQL\Buffers\EntityBuffer;
 use Drupal\graphql\GraphQL\Execution\FieldContext;
 use Drupal\graphql\Plugin\GraphQL\DataProducer\DataProducerPluginBase;
 use Drupal\layout_builder_fieldblock_test\ContextProvider\FakeViewModeContext;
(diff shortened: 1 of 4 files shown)


packages/composer/amazeelabs/silverback_gutenberg/src/Plugin/GraphQL/DataProducer/EditorBlockMedia.php
```diff
@@ -26,31 +33,122 @@
...
+      }
+
+      $context->addCacheableDependency($entity);
+      if (isset($bundles) && !in_array($entity->bundle(), $bundles)) {
+        // If the entity is not among the allowed bundles, don't return it.
+        return NULL;
+      }
+
+      // Get the correct translation.
+      if (isset($language) && $language !== $entity->language()->getId() && $entity instanceof TranslatableInterface) {
```
Leksat: `$entity->hasTranslation($language)` check is missing. `getTranslation` can throw without the check.

packages/composer/amazeelabs/silverback_gutenberg/src/Plugin/GraphQL/DataProducer/EditorBlockMedia.php
```diff
@@ -26,31 +33,122 @@
...
-    $media = Media::load($mediaId);
-    // Try to get the embedded item in the current language or use whatever
-    // we get as a fallback.
-    if (!$media) {
-      \Drupal::logger('graphql_gutenberg')->notice("Cannot load media by ID '{$mediaId}'");
+  public function resolve($block, FieldContext $context): ?Deferred {
+    $id = $block['attrs']['mediaEntityIds'][0] ?? NULL;
+
+    if (!$id) {
       return NULL;
```
Leksat: From PR description:
> Turns out that `$builder->compose` only aborts on `NULL` values when they are emitted by a `Deferred` result.

But here `NULL` is still not wrapped with `Deferred`. Should it be wrapped as well?

Leksat (changes requested):
Just a couple of things 😁

PR #1393, opened 2023-07-24
Style publisher dropdown for Claro theme

Leksat:
## Package(s) involved

`amazeelabs/silverback_publisher_monitor`

### DIFF
--- a/packages/composer/amazeelabs/silverback_publisher_monitor/css/indicator.css
+++ b/packages/composer/amazeelabs/silverback_publisher_monitor/css/indicator.css
@@ -6,7 +6,7 @@
   z-index: 1000;
   position: relative;
   top: 3px;
-  right: 1px;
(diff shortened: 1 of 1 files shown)


Leksat:
@pmelab I don't know why, but if I enable Claro theme in `apps/silverback-drupal`, the dropdown looks different from what I see in silverback-template 🤷 

Also, as we are going to deprecate `apps` in silverback-mono anyway, I'd recommend to style the dropdown in silverback-template environment.

Locally I was linking silverback-mono packages into silverback-template:
```
    {
      "type": "path",
      "url": "../../../silverback-mono/packages/composer/*/*"
    },
```
And then: `composer require amazeelabs/silverback_publisher_monitor:@dev`

(BTW, I think we can use this approach for the future combined testing of the two repos)

PR #1399, opened 2023-08-17
feat(gatsby-fragments): delete fragments before generating

colorfield:
## Package(s) involved

`gatsby-fragments`

## Description of changes

In several commits

- prettier
- add some test files and development docs for manual testing
- delete fragments before generating

Follow-up of https://github.com/AmazeeLabs/silverback-mono/pull/1097

## Motivation and context

When switching tickets on client projects, it happens quite often that we need to run codegen / schema update / fragment generation, because the schema is changing. In this case, `.gql` fragments related to the features are being deleted but the corresponding `.fragment.ts` are left behind. Then this produces local build errors.

A simple way to prevent that is to delete before generating.

## How has this been tested?

Unit tests are green, manually.

### DIFF
--- a/packages/npm/@amazeelabs/gatsby-fragments/.gitignore
+++ b/packages/npm/@amazeelabs/gatsby-fragments/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 index.cjs
+tests/fragments/*.fragment.ts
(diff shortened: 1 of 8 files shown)


colorfield:
We did a peer review session with @Leksat good to merge.
Will only affect legacy projects as the latest version of `silverback-template` doesn't require it anymore.

PR #1402, opened 2023-08-24
chore: introduce "view publisher status" permission

Leksat:
## Package(s) involved

- `silverback_gatsby`
- `silverback_publisher_monitor`

## Motivation and context

To make it possible to completely hide Publisher from toolbar.

## How has this been tested?

Manually.


colorfield:
Thanks @Leksat!

Issue #1274, opened 2023-01-20
Trigger full build automatically in certain cases


isara89:
**Background:**
If an entity is saved that is referenced on multiple nodes, it needs a full build to update the frontend as the change isn't on the respective nodes themselves and therefore the incremental build won't be aware of the changes. 
The same happens e.g. when the menu is updated it also needs a full build as it's used on all pages. 

Examples of entities:
- Menu #1284 
- Paragraph library items / Re-usable Gutenberg blocks #1285
- Media -> to be reproduced/tested

**Ideas:** 
- Trigger a full build once a day
- Specify entities which require a full build and if one or multiple of those are updated, trigger a full build on the next run
- Flag all content where the respective entity is in use as "changed" so they get rebuilt in an incremental build

isara89:
Example of issue with Media: https://amazeelabs.slack.com/archives/C02SVEW2X8F/p1675080894993719
This was fixed after running a full build.

pmelab:
This problem also seems to appear in other projects in more complicated reference-structures. To be reproduced.

PR #1431, opened 2023-10-10
Gatsby references

Leksat:
## Package(s) involved

## Description of changes

Whenever we track an update on an entity, get all its usages and track the source entities as well.

## Motivation and context

`gatsby-graphql-toolkit` creates node references in Gatsby data layer. Yet it does not clean them up when a target node is removed.

```
    if (node) {
      context.gatsbyApi.actions.deleteNode(node)
    }
    // TODO: find all nodes referencing deleted nodes and update/refetch all of them?
```
https://github.com/gatsbyjs/gatsby-graphql-toolkit/blob/cec93fe2457aa0de64dbdeb759626010d9fe270a/src/source-nodes/node-actions/delete-nodes.ts#L17-L20

## Related Issue(s)

https://github.com/AmazeeLabs/silverback-mono/issues/1274

## How has this been tested?

Manually. Added an integration test case.

### DIFF
--- a/packages/tests/silverback-gatsby/playwright-tests/references.spec.ts
+++ b/packages/tests/silverback-gatsby/playwright-tests/references.spec.ts
@@ -1,0 +1,37 @@
+import {
+  drupal,
+  drupalLogin,
+  gatsby,
+  resetState,
+  waitForGatsby,
+} from '@amazeelabs/silverback-playwright';
+import { expect, test } from '@playwright/test';
+
+test.beforeAll(async () => {
+  await resetState();
+});
+
+test('@gatsby-build references', async ({ page }) => {
+  // Check the tag present on the page.
+  await page.goto(gatsby.baseUrl);
+  await page.click('a:text-is("With everything")');
+  await expect(page.locator(':text("Tag 1")')).toHaveCount(1);
+
+  // Remove the tag.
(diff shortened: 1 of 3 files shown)


packages/tests/silverback-gatsby/playwright-tests/references.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+  resetState,
+  waitForGatsby,
+} from '@amazeelabs/silverback-playwright';
+import { expect, test } from '@playwright/test';
+
+test.beforeAll(async () => {
+  await resetState();
+});
+
+test('@gatsby-build references', async ({ page }) => {
```
Leksat, line 14: It should have fallen on the test commit, yet it did not. I don't know why.

Locally it fails properly. But in CI an incremental Gatsby build sources _all_ nodes for some reason.

Another evidence that we need to get rid of `@amazeelabs/silverback-playwright` ASAP. Will do in a separate PR.

pmelab (approved):
So its not going to recursively update chains, right? Just the direct parent level. Should be good enough for the Gutenberg use case.

Leksat:
Yes, just direct parents.

Still I'm a little concerned about the performance. What if we get an entity used in 1k other entities 🙈 Time will tell 😅

PR #1432, opened 2023-10-10
Gatsby `opacity: 0`

pmelab:
## Package(s) involved

`@amazeelabs/gatsby-source-silverback`

## Description of changes

Change Drupal/Gatsby integration to a more schema-first approach.

- Auto-generated fields used internally are prefixed with "\_" so they don't collide with the custom schema.
- Auto-generated fields used internally are marked as `@deprecated` so they are not used in projects directly.
- Gatsby uses UUID's when fetching entities from Drupal.

## Motivation and context

Emphasize explicit custom schemas and make Gatsby "invisible" (hence the witty  PR title).
Projects should not rely magic GraphQL fields and types. Those are an implementation detail by the Drupal/Gatsby integration now. The projects schema, operations and fragments should be valid no matter the implementation (Gatsby/Drupal).

Benefits:
* Reduce magic that has to be understood
* Project code is less bound to Drupal and Gatsby
* No need for exporting schema from Drupal and Gatsby for Typescript
* Use UUID's for easier integration tests.

One key trigger for this PR was the integration of Decap CMS. It turned out that removing Drupal from a template based project was not a trivial task since the schema had a lot of hard dependencies (e.g. all the `queryX` and `loadX` fields). By making the schema the source of truth, that will become a lot easier.

## How has this been tested?

- [x] Manually
- [x] Unit tests
- [x] Integration tests


pmelab:
Current dependencies on/for this PR:
* development
  * **PR #1432** <a href="https://app.graphite.dev/github/pr/AmazeeLabs/silverback-mono/1432" target="_blank">Graphite</a>  👈
    * **PR #1445** <a href="https://app.graphite.dev/github/pr/AmazeeLabs/silverback-mono/1445" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/AmazeeLabs/silverback-mono/1432?utm_source=stack-comment).

pmelab:
@chindris @Leksat : this one is good to review and a little more controversial 😄, therefore i would like to get more opinions.

The failing integration test works on my machine ™️ . Maybe @Leksat has encountered things that break on github? I'm out of ideas.

apps/silverback-gatsby/src/pages/custom-schema.tsx
```diff
@@ -0,0 +1,25 @@
+import { graphql, PageProps } from 'gatsby';
+import React from 'react';
+
+export const query = graphql`
+  query CustomSchema {
+    pages {
+      id
+      title
+    }
+    page(id: "Page:2:en") {
```
Leksat: This may be out of luck without uuids.

Leksat:
> The failing integration test works on my machine ™️

I've seen it a lot that the default content is created in different order locally vs CI. I guess that's the reason.

In Drupal resolver we have a support for uuids:
https://github.com/AmazeeLabs/silverback-mono/blob/b22387e74243b140dbc6dbb10828545483122c4e/packages/composer/amazeelabs/silverback_gatsby/src/Plugin/GraphQL/DataProducer/FetchEntity.php#L196-L207
We use it in tests a lot.

Yet I'm not sure what to do here. Switch to uuids? 🤔 

UPD: Maybe something like `p:text-is("Load: A page")` -> `p:text-matches("Load: .*")` would be enough? (And a similar change for the list check)

Leksat:
Otherwise, the PR looks great 👏 

IIRC we first wanted to add all magic fields directly to schema. But now we would need to add just `id`. I like that! Less work 😁

pmelab:
> IIRC we first wanted to add all magic fields directly to schema. But now we would need to add just `id`. I like that! Less work 😁

Kind of. If we need the "magic" fields outside of the autogenerated stuff (e.g. a static query), we have to explicitly add them like this:

```graphql
type Query {
  gimmeallthepages: [Page] @gatsbyNodes(type: "Page")
}
```

But I think thats fine.

pmelab:
@Leksat thanks for the pointer! Gatsby/Drupal integration has been switched to UUIDs and the latest commits here deserve another review 😄

Leksat (approved):
Looks awesome!

PR #1436, opened 2023-10-14
Linkit BC check

Leksat:
Check if https://github.com/AmazeeLabs/silverback-mono/pull/1385/commits/f7692d5d4f1877132ff6bca27c1d66a9c85772bf works with older version of linkit.


Leksat:
Okay, it does not.

PR #1441, opened 2023-10-17
Run template tests in monorepo PRs

Leksat:


Leksat:
Current problem
```
> graphql-codegen --verbose && swc ./src -d ./build && tsc --emitDeclarationOnly
...
[SUCCESS] Generate to src/generated/source.ts
[FAILED] Cannot use GraphQLScalarType "Boolean" from another module or realm.
[FAILED] 
[FAILED] Ensure that there is only one instance of "graphql" in the node_modules
[FAILED] directory. If different versions of "graphql" are the dependencies of other
[FAILED] relied on modules, use "resolutions" to ensure only one version is installed.
```

Leksat:
Unfortunately, it did not work. Linking monorepo packages opened a gate to the dependency hell. 
```
[FAILED] Ensure that there is only one instance of "graphql" in the node_modules
[FAILED] directory.
```

I tried to sync `graphql` versions, but it did not work either.

I also tried using `file:` protocol for linking. In this case pnpm treats linked packages as real ones and errors when sees `workspace:*` in dependencies.

Anyway, even if it worked, it would be a very very shaky check.

PR #1449, opened 2023-11-01
ResponsiveImageSharp directive

chindris:
Adds a new directive, responsiveImageSharp, that:
- downloads a remote image file and adds it to the schema using Gatsby's createRemoteFileNode (see https://www.gatsbyjs.com/plugins/gatsby-source-filesystem/?=files#createremotefilenode )
- using the Gatsby sharp plugin it generates the responsive image (src, srcset, width, heigth) that can be used to display responsive images, similar to the already existing responsiveImage directive that uses the cloudinary service to resize / crop images. While the cloudinary based directive is able to also take a 'transform' parameter that can be used to customize the end image in a large number of ways (adding borders, rotating, add transparency, etc), the implementation using the Gatsby sharp plugin would just resize / crop images and use the entropy crop focus by default: https://www.gatsbyjs.com/plugins/gatsby-plugin-sharp/#cropfocus

## Package(s) involved
- gatsby-source-silverback
- graphql-directives


apps/silverback-gatsby/gatsby-config.mjs
```diff
@@ -36,7 +37,7 @@ export const plugins = [
       graphql_path: process.env.DRUPAL_GRAPHQL_PATH,
       auth_key: process.env.DRUPAL_AUTH_KEY,
       schema_configuration: './graphqlrc.yml',
-      directives: { echo, responsiveImage },
+      directives: { echo, responsiveImage, responsiveImageSharp },
```
pmelab: I think it should be enough to just pass `responsiveImage: responsiveImageSharp` here? Then there are no changes to the schema necessary.

packages/npm/@amazeelabs/gatsby-source-silverback/directives.graphql
```diff
@@ -2,3 +2,8 @@ directive @sourceFrom(fn: String) on OBJECT
 directive @resolveBy(fn: String) repeatable on FIELD_DEFINITION
 directive @gatsbyNodes(type: String) on FIELD_DEFINITION
 directive @gatsbyNode(type: String, id: String) on FIELD_DEFINITION
+directive @responsiveImageSharp(
```
pmelab: See above. The separate directive should not be necessary. We just provide a different implementation for `@responsiveImage`.

packages/npm/@amazeelabs/graphql-directives/src/resolvers.ts
```diff
@@ -5,6 +5,12 @@ import { flow, isString } from 'lodash-es';
 export function createResolveConfig(
   schema: GraphQLSchema,
   directives: Record<string, GraphQLFieldResolver<any, any>>,
+  gatsbyHelpers?: {
```
pmelab: A nitpick: I would just call this `helpers`, `framework` or `api` and make it completely untyped. There is no benefit in knowing _what_ we pass through here, so we can as well keep it framework-independent.

packages/npm/@amazeelabs/graphql-directives/src/resolvers.ts
```diff
@@ -80,6 +92,7 @@ export function buildResolver(
             processDirectiveArguments(value, args, spec),
             context,
             info,
+            gatsbyHelpers
```
pmelab: I would merge the helper API into the `context`. Thats also where other Gatsby API's like `nodeModel` live. Then we don't have to deviate from the standard GraphQL resolver signature.

pmelab (changes requested):
My mind is blown that this actually works 😄  Thank you! Just a couple of smaller issues we might want to look at.

apps/silverback-gatsby/gatsby-config.mjs
chindris: actually, I kept it like that because in silverback-gatsby we used both, as an example. We use sharp for articles and cloudinary for guternberg pages. But we could change the naming as well. But I changed it now to responsiveImage: responsiveImageSharp

packages/npm/@amazeelabs/gatsby-source-silverback/directives.graphql
chindris: Done that now. I think we will probably have to rename the 'transform' parameter to something else, so that we could also use it as an 'options' parameter for other services. Right now we do not use it for sharp, but we might in the future.

packages/npm/@amazeelabs/graphql-directives/src/resolvers.ts
chindris: Done, renamed to helpers.

packages/npm/@amazeelabs/graphql-directives/src/resolvers.ts
chindris: done, the api is now merged into context. And to make sure they not conflict with something there, those api functions are put inside an api object in the context, so context.api.createNode

packages/npm/@amazeelabs/graphql-directives/src/resolvers.ts
chindris: actually, I renamed them to api now

pmelab (approved):
🥳

PR #1460, opened 2023-11-22
fix: support fragments in Link component

nbaosullivan:
## Package(s) involved

- @amazeelabs/scalars

## Description of changes

Support fragments in links

## Motivation and context

fragments were stripped from URLs when using query-string parseUrl, as parseFragmentIdentifier needs to be set to tru in the function options.

## How has this been tested?

Locally and with vitest


pmelab (approved):
Awesome! Thanks!

PR #1466, opened 2023-12-01
fix: styling when the permission is set to status only

colorfield:
## Package(s) involved

`silverback_publisher_monitor`

## Description of changes

Add extra styles when the drop button is not rendered.

## Motivation and context

If the status permission is the only one to be set, the drop button is not rendered.

## How has this been tested?

Manually, on a project.


pmelab (approved):
Good catch!

PR #1478, opened 2024-01-10
SLB-202: inline fragments in operations

pmelab:
## Package(s) involved

`@amazeelabs/codegen-operation-ids`

## Description of changes

Instead of appending fragments to each operation, they are inlined now.

Before:
```graphql
query {
  ...MyFragment
}

fragment MyFragment on Query {
  myProp
}
```

After:
```graphql
query {
  myProp
}
```

## Motivation and context

Gatsby breaks if it encounters multiple fragments. This change allows us to use the same queries in Gatsby directly, without having duplicate them there.

## Related Issue(s)

[SLB-202](https://amazeelabs.atlassian.net/browse/SLB-202)

## How has this been tested?

- [x] unit tests

[SLB-202]: https://amazeelabs.atlassian.net/browse/SLB-202?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/npm/@amazeelabs/codegen-operation-ids/src/index.test.ts
+++ b/packages/npm/@amazeelabs/codegen-operation-ids/src/index.test.ts
@@ -135,8 +135,8 @@ describe('mode: map', () => {
     ]);
     expect(JSON.parse(result)).toMatchInlineSnapshot(`
       {
-        "Home:37d40553a898c4026ba372c8f42af3df9c3451953b65695b823a8e1e7b5fd90d": "query Home {
-        loadPage(path: \\"/\\") {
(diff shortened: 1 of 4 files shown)


Leksat (approved):
Do I get it right that this can slightly increase the query size, but nothing dramatic?

pmelab:
> Do I get it right that this can slightly increase the query size, but nothing dramatic?

Yes it will increase query size. We don't have a lot of fragments that are re-used within the same query. The most notable example is `MediaImage`, and that one is rather small. I don't think this is an issue because:

* Queries are never sent over the wire. They are persisted and the HTTP request only contains the ID.
* The GraphQL parser is pretty fast. We have only seen issues with document size when attempting to cache the whole AST in Drupal.
* Execution stays the same.

PR #1488, opened 2024-02-06
feat(SLB-255): allow primitive values in @value directive

pmelab:
## Package(s) involved

`@amazeelabs/graphql_directives`

## Description of changes

Add additional `int`, `float`, `string` and `boolean` parameters to the `@value` directive.

## Motivation and context

A little nicer GraphQL code when we don't have to encode objects.

## Related Issue(s)

SLB-255

## How has this been tested?

* Unit tests

### DIFF
--- a/packages/composer/amazeelabs/graphql_directives/tests/src/Kernel/UtilitiesTest.php
+++ b/packages/composer/amazeelabs/graphql_directives/tests/src/Kernel/UtilitiesTest.php
@@ -24,6 +24,13 @@ class UtilitiesTest extends GraphQLTestBase {
     $this->assertResults('{ foo }', [], ['foo' => 'bar']);
   }
 
+  function testValuePrimitives() {
+    $this->assertResults('{ primitiveString }', [], ['primitiveString' => 'foo']);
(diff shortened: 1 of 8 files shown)


packages/composer/amazeelabs/graphql_directives/tests/src/Kernel/UtilitiesTest.php
```diff
@@ -24,6 +24,13 @@ function testSchemaLoading() {
     $this->assertResults('{ foo }', [], ['foo' => 'bar']);
   }
 
+  function testValuePrimitives() {
+    $this->assertResults('{ primitiveString }', [], ['primitiveString' => 'foo']);
+    $this->assertResults('{ primitiveInt }', [], ['primitiveInt' => 42]);
+    $this->assertResults('{ primitiveFloat }', [], ['primitiveFloat' => 3.14]);
+    $this->assertResults('{ primitiveBoolean }', [], ['primitiveBoolean' => true]);
```
Leksat, line 31: I'd also add a case for `null`.

But I'm already happy that it is possible to write just `@value` instead of `@value(json: "null")` 🎉

Leksat (approved):
Looks good!

PR #1503, opened 2024-03-20
feat(silverback_gatsby): default entity fetch operation

colorfield:
## Package(s) involved

`silverback_gatsby`

## Description of changes

Delegate to the `fetch_entity` data producer default `access_operation` value if `operation` is not set.

## Motivation and context

Directives can be simplified as

```graphql
previewPage(id: ID!, rid: ID, locale: String!): Page
    @fetchEntity(type: "node", id: "$id", rid: "$rid", language: "$locale")
```

instead of

```graphql
previewPage(id: ID!, rid: ID, locale: String!): Page
    @fetchEntity(type: "node", id: "$id", rid: "$rid", language: "$locale", operation: "view")
```

## How has this been tested?

Manually, on a project.

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -15,3 +15,5 @@
 /markdown.xml
 /misc.xml
 /inspectionProfiles/profiles_settings.xml
+# GitHub Copilot persisted chat sessions
+/copilot/chatSessions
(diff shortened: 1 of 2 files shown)


amazee-labs-kitt (approved):
Thanks! 🍾

amazee-labs-kitt:
Good catch to use the default operation instead of just `view`. That would have caught up with us at some point and nobody would have had an idea whats going on 🙄

PR #1507, opened 2024-04-01
Cleanup

Leksat:
Main change: Schema tests are now executed with Vitest

Also removed `@amazeelabs/graphql-schema-test` package as it isn't used anymore


Leksat:
FYI @pmelab

PR #1511, opened 2024-04-03
`@drupalView` directive

Leksat:
## Package(s) involved

`amazeelabs/graphql_directives`

## Related Issue(s)

SLB-221

## How has this been tested?

Schema tests.

### DIFF
--- a/packages/composer/amazeelabs/graphql_directives/src/Plugin/GraphQL/DataProducer/DrupalView.php
+++ b/packages/composer/amazeelabs/graphql_directives/src/Plugin/GraphQL/DataProducer/DrupalView.php
@@ -1,0 +1,136 @@
+<?php
+
+namespace Drupal\graphql_directives\Plugin\GraphQL\DataProducer;
+
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Entity\TranslatableInterface;
+use Drupal\Core\Render\Element;
+use Drupal\Core\Render\RenderContext;
+use Drupal\graphql\GraphQL\Execution\FieldContext;
+use Drupal\graphql\Plugin\GraphQL\DataProducer\DataProducerPluginBase;
+use Drupal\views\ResultRow;
+use Drupal\views\Views;
+
+/**
+ * @DataProducer(
+ *   id = "drupal_view",
+ *   name = @Translation("Drupal view"),
+ *   description = @Translation("Executes a Drupal view."),
+ *   produces = @ContextDefinition("any",
+ *     label = @Translation("An array with view info and results"),
+ *     multiple = TRUE
+ *   ),
+ *   consumes = {
+ *     "id" = @ContextDefinition("string",
+ *       label = @Translation("View ID and display ID divided by a comma"),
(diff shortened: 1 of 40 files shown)


packages/composer/amazeelabs/graphql_directives/src/Plugin/GraphQL/DataProducer/DrupalView.php
```diff
@@ -0,0 +1,136 @@
...
+          'contextual_filters' => $contextualFilters,
+        ],
+        'rows' => $rows,
+        'total' => $total,
+        'filters' => $filters,
+      ];
+    });
+  }
+
+  function withCacheContext(FieldContext $resolveContext, callable $callback) {
```
pmelab, line 125: Sometimes I wonder if the GraphQL module should do that by default 😅

packages/composer/amazeelabs/graphql_directives/src/Plugin/GraphQL/DataProducer/DrupalView.php
```diff
@@ -0,0 +1,136 @@
...
+ *   description = @Translation("Executes a Drupal view."),
+ *   produces = @ContextDefinition("any",
+ *     label = @Translation("An array with view info and results"),
+ *     multiple = TRUE
+ *   ),
+ *   consumes = {
+ *     "id" = @ContextDefinition("string",
+ *       label = @Translation("View ID and display ID divided by a comma"),
+ *     ),
+ *     "args" = @ContextDefinition("string",
```
pmelab, line 27: We ended up with the same approach in https://github.com/AmazeeLabs/silverback-template/pull/147

Discussed with @chindris that deriving GraphQL types from Drupal config (e.g. views filters, webform inputs) could be interesting. Just as a reference.

pmelab (approved):
Can't come up with anything to complain about 😬

packages/composer/amazeelabs/graphql_directives/src/Plugin/GraphQL/DataProducer/DrupalView.php
Leksat: There's an ultimate solution: https://www.drupal.org/project/drupal/issues/2638686#comment-13818152

packages/composer/amazeelabs/graphql_directives/src/Plugin/GraphQL/DataProducer/DrupalView.php
Leksat: A little sad thing is that a tiny typo can break it unnoticeably. For example, if `tag` filter is multi-value, it should be passed as `tag[]=1`. If it's sent as `tag=1`, the value will be ignored.

Checking every possible filter with e2e tests does not sound like a great idea 😕

Leksat: Probably backend devs should provide TypeScript types for `args` 🤷

PR #1539, opened 2024-06-26
feat(SLB-352): waku bridge

pmelab:
## Package(s) involved

`@amazeelabs/bridge-waku`

## Description of changes

`@amazeelabs/bridge` implementation for [waku](https://waku.gg)

## Related Issue(s)

[SLB-352](https://amazeelabs.atlassian.net/browse/SLB-352)

## How has this been tested?

In a `silverback-template` PR.

[SLB-352]: https://amazeelabs.atlassian.net/browse/SLB-352?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/npm/@amazeelabs/bridge-waku/src/client.tsx
```diff
@@ -0,0 +1,33 @@
...
+export const createUseLocationHook: (
+  useRouter: typeof useRouter_UNSTABLE,
+) => useLocationType = (useRouter) => () => {
+  const router = useRouter();
+
+  // TODO: Use router.hash when https://github.com/dai-shi/waku/pull/746 is merged.
+  const [hash, setHash] = useState('');
+  useEffect(() => {
+    setHash(window.location.hash);
+  }, [router]);
```
Leksat, line 17: Same question as here https://github.com/dai-shi/waku/pull/746/files#r1645256634 🙂

pmelab: There is a very small chance that `window.location.hash` changes without the router changing, and I think that case would rather be a bug somewhere else 🤔

But I think i would rather fix the Waku-PR and remove this entirely when we can update.

PR #1552, opened 2024-07-12
feat(SLB-416): preview link share api

colorfield:
## Package(s) involved

`silverback_preview_link`

## Description of changes

Adds share api when supported https://caniuse.com/?search=Navigator%20share


colorfield:
Closing we will implement copy to clipboard instead, that has better support and is more relevant EX wise on desktop. See #1553

PR #1574, opened 2024-09-06
Drupal 11

Leksat:


apps/silverback-drupal/web/themes/custom/silverback_iframe_test/silverback_iframe_test.info.yml
```diff
@@ -3,7 +3,7 @@ type: theme
 base theme: silverback_iframe_theme
 description: 'A sub-theme for silverback_iframe_theme.'
 package: Silverback Tests
-core_version_requirement: ^8 || ^9 || ^10
+core_version_requirement: ^8 || ^9 || ^10 || ^11
```
dan2k3k4, line 6: Do we really need `8+` and `9+` ?

Also, wouldn't something like `>8 || <12` work to do the same thing?

Leksat: No, yes 🙂 

Keeping 8 and 9 does not hurt. Using `^` is more common in Drupal world.

Leksat:
GraphQL has no release for D11. Webform is not ready for D11. So I will stop for now.

PR #1596, opened 2024-11-22
feat: prevent .envrc from overriding any changes you make.

elistone:
## Package(s) involved
- sliverback-cli

## Description of changes
Adding `override: false` on the scaffolding of `.envrc`

## Motivation and context
If I want to make changes locally for example configure `devbox` I don't want my changes to be overiden every time I run `composer install`.


elistone:
new PR #1597

PR #1612, opened 2025-02-04
Autosave fixes

Leksat:
## Package(s) involved

`amazeelabs/silverback_autosave`

## Description of changes

1. Add a serial ID field and use it as a primary key - to avoid the duplicate id error
2. Add indexes to the table fields - to make queries faster

## Motivation and context

Random `duplicate entry for key … in silverback_autosave_entity_form` error.

## Related Issue(s)

SLB-511

## How has this been tested?

Manually on the local setup.


pmelab (approved):
Nice! Let's merge and integrate it! (and update in all projects that use it)

